PR #18, opened 2019-11-23
api: Remove skip_decoding option

michi-covalent:
Now that flows always get decoded, we don't need this option anymore.


michi-covalent:
2 commits:
- remove skip_decoding option
- clean up filters package to not use payload

pkg/filters/filters.go
```diff
@@ -282,17 +282,8 @@ func filterByVerdicts(vs []pb.Verdict) FilterFunc {
 func filterByEventType(types []*pb.EventTypeFilter) FilterFunc {
 	return func(ev *v1.Event) bool {
 		event := ev.GetFlow().GetEventType()
-		// fall back on payload data for undecoded types
```
gandro: This commit misses the "Signed-off-By"

michi-covalent: done

PR #30, opened 2019-12-05
 api: Introduce interface for pb.Flow 

gandro:
This introduces an interface for `pb.Flow` and generalizes `v1.Event` to return an instance of the interface. This allows better mocking and alternative flow implementations in the future.

The second PR moves the `metrics.ProcessFlow` over to the new interface .

### DIFF
--- a/pkg/api/v1/interface.go
+++ b/pkg/api/v1/interface.go
@@ -1,0 +1,49 @@
+// Copyright 2019 Authors of Hubble
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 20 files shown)


pkg/api/v1/interface.go
```diff
@@ -0,0 +1,49 @@
...
+	GetL7() *pb.Layer7
+	GetReply() bool
+	GetEventType() *pb.CiliumEventType
+	GetSourceService() *pb.Service
+	GetDestinationService() *pb.Service
+	GetSummary() string
+}
+
+// This ensures that the protobuf definition implements the interface
+var _ Flow = &pb.Flow{}
```
michi-covalent: neat, although i kind of wish we can just auto-generate this interface.

gandro: Agreed. Unfortunately, I haven't found a tool that nicely implements the interface that without all the `XXX_` methods created by protoc-gen-go.

michi-covalent:
ok merging

PR #33, opened 2019-12-06
filter: Add support for filtering by reply field

michi-covalent:
The filter is defined as a repeated field even though it's a boolean
filter so that we can distinguish the filter being explicity set to
false from the zero value of boolean type.


pkg/filters/filters.go
```diff
@@ -465,6 +465,24 @@ func filterByPort(portStrs []string, getPort func(*v1.Event) (port uint16, ok bo
 	}, nil
 }
 
+func filterByReplyField(replyParams []bool) FilterFunc {
+	return func(ev *v1.Event) bool {
+		if len(replyParams) == 0 {
+			return true
+		}
+		if ev.GetFlow() == nil {
```
gandro: Unfortunately, since #30 `GetFlow` returns an interface type and this check will never be true. We should probably add a `HasFlow()` to `v1.Event` for this purpose.

michi-covalent: rebased with #30 merged. added a type switch for `ev.Event.(type)` instead of checking if `GetFlow` is nil.

PR #34, opened 2019-12-10
Payload handling

glibsm:
By default the entire binary payload was sent over the wire, which is duplicate info.

❯ git log origin/master..HEAD --oneline --no-decorate
103c29b pkg/server: Remove binary payload by default
4ff3dab api/v1: Boolean flag for payload


glibsm:
closing in favor of the newer PR

PR #36, opened 2019-12-11
Add support for specifying server in client via env variable

joestringer:
Allow the $HUBBLE_SOCK environment variable to override the server socket
path for the client.


glibsm:
what takes precedence if both are specified?

joestringer:
I believe that explicit CLI arg that is provided will take precedence over env variables.

PR #47, opened 2020-01-10
helm: set priorityClassName to system-node-critical for hubble pods

### LINKED ISSUE #25: scheduler.alpha.kubernetes.io/critical-pod is obsolete in k8s 1.16
According to https://github.com/kubernetes/kubernetes/blob/e9ce31bcb3feb411c4de4902651e43dbb78ed151/CHANGELOG-1.16.md, the annotation scheduler.alpha.kubernetes.io/critical-pod is obsolete, and Pod priority (spec.priorityClassName) should be used instead to mark pods as critical.

Rolinh:
The annnotation `scheduler.alpha.kubernetes.io/critical-pod` is now obsolete and has to be replaced by the priorityClassName spec instead to mark an add-on pod as critical.

There exists 2 priority classes to mark an add-on pod as critical (only under the `kube-system` namespace), namely:
- `system-cluster-critical` (value: 2000000000)
- `system-node-critical`    (value: 2000001000)

As the hubble pod is critical to the node itself, the system-node-critical priority class is appropriate for hubble pods.

Closes #25

### DIFF
--- a/install/kubernetes/hubble/templates/daemonset.yaml
+++ b/install/kubernetes/hubble/templates/daemonset.yaml
@@ -10,13 +10,8 @@ spec:
       kubernetes.io/cluster-service: "true"
   template:
     metadata:
-      annotations:
-        # This annotation plus the CriticalAddonsOnly toleration makes
-        # hubble to be a critical pod in the cluster, which ensures hubble
-        # gets priority scheduling, same as cilium-agent.
-        # https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/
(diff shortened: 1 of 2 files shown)


Rolinh:
> Looks good to me. In Cilium we have a check to only enable this on Kubernetes 1.11+, but since 1.10 is EOL at this point and Cilium also drops support in 1.7 as well, I think it's fine to leave it.

The check in cilium also ensures that this setting is only applied when the namespace is `kube-system`, which is relevant here as well. However, in a case where hubble wouldn't run in the `kube-system` namespace, a custom `priorityClass` shall be defined to set the pod priority. This should probably be done in a separate PR I think.

PR #49, opened 2020-01-13
pkg/printer: Use interface in print methods

glibsm:
Signed-off-by: Glib Smaga <code@gsmaga.com>

### DIFF
--- a/pkg/printer/printer.go
+++ b/pkg/printer/printer.go
@@ -21,10 +21,10 @@ import (
 	"strings"
 	"text/tabwriter"
 
-	pb "github.com/cilium/hubble/api/v1/flow"
-	"github.com/cilium/hubble/pkg/format"
-
 	"github.com/cilium/cilium/pkg/monitor/api"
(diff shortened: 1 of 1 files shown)


pkg/printer/printer.go
```diff
@@ -157,14 +158,14 @@ func getFlowType(f *pb.Flow) string {
...
-	if f.Verdict == pb.Verdict_DROPPED {
-		return api.DropReason(uint8(f.EventType.SubType))
+	if f.GetVerdict() == pb.Verdict_DROPPED {
+		return api.DropReason(uint8(f.GetEventType().SubType))
 	}
-	return api.TraceObservationPoint(uint8(f.EventType.SubType))
+	return api.TraceObservationPoint(uint8(f.GetEventType().SubType))
 }
 
 // WriteProtoFlow writes pb.Flow into the output writer.
```
Rolinh: I think the method comment should be updated: `s/pb.Flow/f`

PR #55, opened 2020-01-20
refactor: Move ipCacheEvents channel inside ObserverServer

michi-covalent:
To be consistent with other channels like endpointEvents and logRecord.

### DIFF
--- a/pkg/server/ipcache.go
+++ b/pkg/server/ipcache.go
@@ -103,9 +103,14 @@ func (s *ObserverServer) processIPCacheEvent(an monitorAPI.AgentNotify) bool {
 	return false
 }
 
-// syncIPCache initializes the IPCache by fetching an initial version from
-// Cilium and then starts reading IPCacheNotification from the channel.
-func (s *ObserverServer) syncIPCache(ipcacheEvents <-chan monitorAPI.AgentNotify) {
+// syncIPCache will obtain an initial IPCache snapshot from Cilium
+// and then start mirroring IPCache events based on IPCacheNotification sent
+// through the ipCacheEvents channels. Only messages of type
+// `AgentNotifyIPCacheUpserted` and `AgentNotifyIPCacheDeleted` should be sent
+// through that channel. This function assumes that the caller is already
+// connected to Cilium Monitor, i.e. no IPCacheNotification must be lost after
+// calling this method.
+func (s *ObserverServer) syncIPCache() {
 	for {
 		err := s.fetchIPCache()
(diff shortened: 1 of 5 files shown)


pkg/server/ipcache.go
```diff
@@ -103,9 +103,14 @@ func (s *ObserverServer) processIPCacheEvent(an monitorAPI.AgentNotify) bool {
...
-// Cilium and then starts reading IPCacheNotification from the channel.
-func (s *ObserverServer) syncIPCache(ipcacheEvents <-chan monitorAPI.AgentNotify) {
+// syncIPCache will obtain an initial IPCache snapshot from Cilium
+// and then start mirroring IPCache events based on IPCacheNotification sent
+// through the ipCacheEvents channels. Only messages of type
+// `AgentNotifyIPCacheUpserted` and `AgentNotifyIPCacheDeleted` should be sent
+// through that channel. This function assumes that the caller is already
+// connected to Cilium Monitor, i.e. no IPCacheNotification must be lost after
+// calling this method.
+func (s *ObserverServer) syncIPCache() {
```
gandro: While I agree that the current interface is ugly, it was created to enforce this constraint ("This function assumes that the caller is already connected to Cilium Monitor"). Unfortunately with this PR, this constraint can easily be broken:

In cases where it takes Hubble a while to open the Cilium Monitor socket, we will end up losing updates:
1. `syncIPCache()` is called in `Start()`, an initial ipcache is fetched.
2. Cilium send upsert event, since nobody is connected to the socket, it is lost.
3. `consumeMonitorEvents()` is started, only then the monitor socket is opened, but now the event is already lost.

michi-covalent: do we have similar issues with syncing endpoints / fqdn, or is this unique to ipcache?

gandro: I think the other caches currently run into the danger of loosing events, though I think some cases they are able to recover, where as IP cache is not. For example endpoint events are always indexed by the endpoint ID which never changes. Meanwhile, the CIlium IP cache implementation has logic for "key change events" when an IP is promoted to a CIDR. If this happens and we miss that update, any subsequent updates for that IP will fail (since we missed the key update). I'm not entirely sure when/if this happens, to be honest, and if it is a problem in practise. 

I do think however that we should unify the way we update caches. I would highly prefer for the endpoint cache to work similarly to IP cache, i.e. having a single channel for add/remove events, instead of having two (which introduces the danger of events arriving out of order).

PR #58, opened 2020-01-20
serve: Add "local" mode

michi-covalent:
LocalObserverServer that simply ignores all the state change events. It's
meant to be started embeeded in Cilium process, and thus the parser can
access the Cilium state directly without maintaining a copy.

Note that LocalObserverServer still connects to the monitor socket to
receive monitor events from Cilium. We need to figure out how to access
monitor events from Cilium's ring buffer.

### DIFF
--- a/cmd/serve/serve.go
+++ b/cmd/serve/serve.go
@@ -90,7 +90,30 @@ func New(log *zap.Logger) *cobra.Command {
 				}()
 			}
 
-			err = Serve(log, listenClientUrls)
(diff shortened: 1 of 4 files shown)


michi-covalent:
a rough draft for cilium-side change as a reference: https://github.com/cilium/cilium/tree/pr/michi/hubble

michi-covalent:
not ready for review just yet, need to make some small changes.

PR #59, opened 2020-01-21
serve: Add a "local" server

michi-covalent:
LocalObserverServer. It ignores all the state change events and does not
connect to Cilium's monitor socket. It's meant to be started embeeded in
Cilium process so that:

1. the parser can access Cilium state directly without making a copy, and
2. the server can receive monitor payloads by registering a listener.

### DIFF
--- a/cmd/serve/serve.go
+++ b/cmd/serve/serve.go
@@ -91,7 +91,32 @@ func New(log *zap.Logger) *cobra.Command {
 				}()
 			}
 
-			err = Serve(log, listenClientUrls)
+			ciliumClient, err := client.NewClient()
+			if err != nil {
+				log.Fatal("failed to get Cilium client", zap.Error(err))
+			}
+			ipCache := ipcache.New()
+			fqdnCache := fqdncache.New()
+			serviceCache := servicecache.New()
+			endpoints := v1.NewEndpoints()
+			podGetter := &server.LegacyPodGetter{
+				PodGetter:      ipCache,
+				EndpointGetter: endpoints,
+			}
+			payloadParser, err := parser.New(endpoints, ciliumClient, fqdnCache, podGetter, serviceCache)
+			if err != nil {
+				log.Fatal("failed to get parser", zap.Error(err))
+			}
+			s := server.NewServer(
+				ciliumClient,
+				endpoints,
+				ipCache,
+				fqdnCache,
+				serviceCache,
+				payloadParser,
+				int(maxFlows),
+			)
+			err = Serve(log, listenClientUrls, s)
 			if err != nil {
 				log.Fatal("", zap.Error(err))
 			}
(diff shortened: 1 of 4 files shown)


michi-covalent:
rebased to pick up https://github.com/cilium/hubble/commit/657284b286e34c6528053ec50406de5722a352f8

cmd/serve/serve.go
```diff
@@ -296,6 +292,11 @@ func Serve(log *zap.Logger, listenClientUrls []string) error {
 	setupSigHandler()
 	fmt.Printf("Press Ctrl-C to quit\n")
 
+	if !s.UseMonitorSocket() {
+		// Don't connect to monitor socket. Block here.
+		select {}
```
glibsm: 🤔 

Trying to decide if this signals something bad about this function that it requires noop blocking

michi-covalent: we probably don't need to do this. on hubble side i'm doing https://github.com/cilium/cilium/commit/f0db2ca2e9ef82fb01e1150b2aca4dd2e87274ea#diff-1113406423fe93e2d600aa0f9e69183bR1353

we can call runHubble without creating a new goroutine if Serve() doesn't block i think.

michi-covalent: something like this: https://github.com/cilium/cilium/commit/0c0f38988608d10706dfa9b8e19895c2bcbe4125#diff-1113406423fe93e2d600aa0f9e69183bR1352

pkg/server/observer.go
```diff
@@ -41,6 +41,22 @@ import (
...
 
+// Observer defines the interface for observer server.
+type Observer interface {
+	observer.ObserverServer
+	Start()
+	GetEventsChannel() chan *pb.Payload
+	GetEndpointEventsChannel() chan<- monitorAPI.AgentNotify
+	GetLogRecordNotifyChannel() chan<- monitor.LogRecordNotify
+	StartMirroringIPCache(ipCacheEvents <-chan monitorAPI.AgentNotify)
+	StartMirroringServiceCache(ipCacheEvents <-chan monitorAPI.AgentNotify)
```
gandro: How hard would it be to remove these from the interface? I would love to end up in a situation where we are able to split the observer fully into two parts/interfaces:

- One part contains a read-only reference to the ring buffer and uses it to serve the gRPC API (thus doesn't need these methods)
- The other part reads from cilium monitor to both populate the caches (when running Hubble outside of Cilium in a separate process) and the ring-buffer.

michi-covalent: probably not that hard. i think consumeMonitorEvents is the only thing that calls these methods:  https://github.com/cilium/hubble/commit/174a5a38b421b7c9f638e59fde8d17fdb276a185#diff-2079af2803eabcc08ae613577ea25d47R392 and local server doesn't call consumeMonitorEvents.

let me play around a bit.

Rolinh: I fully agree that the observer is doing too much and should be split into smaller components with less responsibilities. However, I don't think such involved work should be part of this PR. We should maybe open an issue for that one.

michi-covalent: @gandro @Rolinh i opened another pr https://github.com/cilium/hubble/pull/64 . it's slightly different from what @gandro suggested, but i think it's cleaner than this pr.

- clean up the Observer interface: https://github.com/cilium/hubble/pull/64/files#diff-69945783af04950df9270c268c24c6a1R52. it adds a single method HandleMonitorSocket that encapsulates the logic for connecting to monitor socket and consuming monitor events. it's a noop for local server.
- move all the code related to monitor socket handling from cmd/serve.go to server/observer.go: https://github.com/cilium/hubble/pull/64/files#diff-69945783af04950df9270c268c24c6a1R468. cmd/serve.go simply calls HandleMonitorSocket.

michi-covalent:
moved to https://github.com/cilium/hubble/pull/64

PR #60, opened 2020-01-22
enable service address to service name translation

### LINKED ISSUE #48: Translate service IPs to service names
Similar to how we map IP addresses to domain and pod names, we should also map service IPs to the Kubernetes service name. The necessary changes in Cilium have already been done ( https://github.com/cilium/cilium/pull/9554 https://github.com/cilium/cilium/pull/9574), what still remains are the necessary changes in Hubble.

Suggested steps:

1. Extend our [Cilium client](https://github.com/cilium/hubble/blob/299831f3608cdcff0ebc79697d0a645f168e5ead/pkg/cilium/client/client.go) to fetch the services from the Cilium API
2. Initialize a service cache from the Cilium API (`/v1/service/`)
3. Parse `AgentNotifyService{Upserted,Deleted}` messages (similarly to how IPCache notifications work, see [serve.go](https://github.com/cilium/hubble/blob/299831f3608cdcff0ebc79697d0a645f168e5ead/cmd/serve/serve.go#L390) and [ipcache.go](https://github.com/cilium/hubble/blob/299831f3608cdcff0ebc79697d0a645f168e5ead/pkg/server/ipcache.go)), in-cooperate updates into the service cache
4. Use the service cache in both the `threefour` and `seven` parser to populate the `SourceService` or `DestinationService` field
5. Extend [printer](https://github.com/cilium/hubble/blob/299831f3608cdcff0ebc79697d0a645f168e5ead/pkg/printer/printer.go#L113) to display the service name in the CLI

Rolinh:
This turns the output of `hubble observe` from this:
```
Jan 22 13:25:22.808 []: default/alpine:55822 -> 172.20.0.113:80(http) from-endpoint FORWARDED (TCP Flags: SYN)
```
To this (when ip translation is enabled):
```
Jan 22 13:25:22.808 []: default/alpine:55822 -> default/deathstar:80(http) from-endpoint FORWARDED (TCP Flags: SYN)
```
In short, the service IP is replaced by `<namespace>/<service name>`. The service port is left as it can be a useful information for whoever is observing the flow.

Note that for this feature to work, Cilium 1.7+ must be used. With Cilium <= 1.6, service addresses are simply not translated.

Closes #48

### DIFF
--- a/cmd/serve/serve.go
+++ b/cmd/serve/serve.go
@@ -45,6 +45,7 @@ import (
 	metricsAPI "github.com/cilium/hubble/pkg/metrics/api"
 	"github.com/cilium/hubble/pkg/parser"
 	"github.com/cilium/hubble/pkg/server"
+	"github.com/cilium/hubble/pkg/servicecache"
 	"github.com/gogo/protobuf/types"
 	"github.com/google/gops/agent"
 	"github.com/spf13/cobra"
@@ -237,13 +238,14 @@ func Serve(log *zap.Logger, listenClientUrls []string) error {
 
 	ipCache := ipcache.New()
 	fqdnCache := fqdncache.New()
+	serviceCache := servicecache.New()
 	endpoints := v1.NewEndpoints()
 	podGetter := &server.LegacyPodGetter{
 		PodGetter:      ipCache,
 		EndpointGetter: endpoints,
 	}
 
-	payloadParser, err := parser.New(endpoints, ciliumClient, fqdnCache, podGetter)
+	payloadParser, err := parser.New(endpoints, ciliumClient, fqdnCache, podGetter, serviceCache)
 	if err != nil {
 		return err
 	}
@@ -253,6 +255,7 @@ func Serve(log *zap.Logger, listenClientUrls []string) error {
 		endpoints,
 		ipCache,
 		fqdnCache,
+		serviceCache,
 		payloadParser,
 		int(maxFlows),
 	)
@@ -397,6 +400,9 @@ func consumeMonitorEvents(s *server.ObserverServer, conn net.Conn, version liste
 	ipCacheEvents := make(chan monitorAPI.AgentNotify, 100)
 	s.StartMirroringIPCache(ipCacheEvents)
 
+	serviceEvents := make(chan monitorAPI.AgentNotify, 100)
+	s.StartMirroringServiceCache(serviceEvents)
+
 	getParsedPayload, err := getMonitorParser(conn, version)
 	if err != nil {
 		return err
@@ -426,10 +432,12 @@ func consumeMonitorEvents(s *server.ObserverServer, conn net.Conn, version liste
 				monitorAPI.AgentNotifyEndpointRegenerateSuccess,
 				monitorAPI.AgentNotifyEndpointDeleted:
 				endpointEvents <- an
-			case monitorAPI.AgentNotifyIPCacheUpserted:
+			case monitorAPI.AgentNotifyIPCacheUpserted,
+				monitorAPI.AgentNotifyIPCacheDeleted:
 				ipCacheEvents <- an
-			case monitorAPI.AgentNotifyIPCacheDeleted:
-				ipCacheEvents <- an
+			case monitorAPI.AgentNotifyServiceUpserted,
+				monitorAPI.AgentNotifyServiceDeleted:
+				serviceEvents <- an
 			}
 		case monitorAPI.MessageTypeAccessLog:
 			// TODO re-think the way this is being done. We are dissecting/
--- a/pkg/parser/getters/getters.go
+++ b/pkg/parser/getters/getters.go
@@ -17,9 +17,11 @@ package getters
 import (
 	"net"
 
-	"github.com/cilium/cilium/api/v1/models"
+	pb "github.com/cilium/hubble/api/v1/flow"
 	v1 "github.com/cilium/hubble/pkg/api/v1"
 	"github.com/cilium/hubble/pkg/ipcache"
+
+	"github.com/cilium/cilium/api/v1/models"
 )
 
 // DNSGetter ...
@@ -45,4 +47,10 @@ type IdentityGetter interface {
 type IPGetter interface {
 	// GetIPIdentity fetches information known about a remote IP.
 	GetIPIdentity(ip net.IP) (identity ipcache.IPIdentity, ok bool)
+}
+
+// ServiceGetter fetches service metadata.
+type ServiceGetter interface {
+	GetServiceByAddr(ip net.IP, port uint16) (service pb.Service, ok bool)
+	GetServiceByID(id int64) (service pb.Service, ok bool)
 }
--- a/pkg/parser/seven/parser.go
+++ b/pkg/parser/seven/parser.go
@@ -39,13 +39,14 @@ import (
 
 // Parser is a parser for L7 payloads
 type Parser struct {
-	cache     *lru.Cache
-	dnsGetter getters.DNSGetter
-	ipGetter  getters.IPGetter
+	cache         *lru.Cache
+	dnsGetter     getters.DNSGetter
+	ipGetter      getters.IPGetter
+	serviceGetter getters.ServiceGetter
 }
 
 // New returns a new L7 parser
-func New(dnsGetter getters.DNSGetter, ipGetter getters.IPGetter, opts ...options.Option) (*Parser, error) {
+func New(dnsGetter getters.DNSGetter, ipGetter getters.IPGetter, serviceGetter getters.ServiceGetter, opts ...options.Option) (*Parser, error) {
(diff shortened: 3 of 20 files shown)


pkg/parser/seven/parser.go
```diff
@@ -123,11 +125,22 @@ func (p *Parser) Decode(payload *pb.Payload, decoded *pb.Flow) error {
 		}
 	}
 
+	l4, sourcePort, destinationPort := decodeLayer4(r.TransportProtocol, sourceEndpoint, destinationEndpoint)
```
gandro: nit: Is there a reason you chose `int` for the port number?

To my knowledge, there is no case where a port would not fit into 16 bits and Cilium always sends a uint16 anyways. I think it would save you a few casts if the ports returned here were `uint16` and `ServiceGetter.GetServiceByAddr()` would take a `uint16` as well.

pkg/parser/threefour/parser.go
```diff
@@ -259,6 +273,7 @@ func decodeLayers(packet *packet) (
 	ip *pb.IP,
 	l4 *pb.Layer4,
 	sourceIP, destinationIP net.IP,
+	sourcePort, destinationPort uint32,
```
gandro: nit: Same as above, I would prefer `uint16`.

pkg/printer/printer_test.go
```diff
@@ -142,27 +142,25 @@ func Test_getHostNames(t *testing.T) {
...
 		f *pb.Flow
 	}
+	type want struct {
+		src, dst string
+	}
 	tests := []struct {
-		name  string
-		args  args
-		want  string
-		want1 string
```
gandro: Nice catch! Thanks for cleaning this up!

pkg/server/service.go
```diff
@@ -0,0 +1,104 @@
...
+			if err := s.fetchServiceCache(); err != nil {
+				s.log.Error("Failed to fetch service cache from Cilium", zap.Error(err))
+			}
+			refresh.Reset(serviceCacheInitRetryInterval)
+		case an, ok := <-serviceEvents:
+			if !ok {
+				return
+			}
+			// Initially we might see stale updates that were enqued before we
+			// initialized the ipcache. Once we see the first applicable update
```
gandro: nit: Comment still mentions `ipcache`.

pkg/server/service.go
```diff
@@ -0,0 +1,104 @@
...
+			// though, all subsequent updates must be applicable as well.
+			updated := s.processServiceEvent(an)
+			switch {
```
gandro: While I added this "is caught up" logic initially originally in `ipcache.go` myself, I'm honestly not sure if it carries its weight, especially for the service cache which has immutable service IDs (which the IP cache apparently doesn't have). Feel free to just remove it if you feel the same.

pkg/servicecache/service_cache.go
```diff
@@ -0,0 +1,174 @@
...
+		cache: map[string]*entry{},
+	}
+}
+
+type entry struct {
+	ID           int64
+	Name         string
+	Namespace    string
+	FrontendIP   net.IP
+	FrontendPort int
```
gandro: nit: Same as above, I think `FrontendPort` can be `uint16`.

pkg/parser/getters/getters.go
```diff
@@ -46,3 +48,9 @@ type IPGetter interface {
 	// GetIPIdentity fetches information known about a remote IP.
 	GetIPIdentity(ip net.IP) (identity ipcache.IPIdentity, ok bool)
 }
+
+// ServiceGetter fetches service metadata.
+type ServiceGetter interface {
+	GetServiceByAddr(ip net.IP, port int) (service pb.Service, ok bool)
+	GetServiceByID(id int64) (service pb.Service, ok bool)
```
gandro: nit: This function seems unused - did you have a future use for in mind?

pkg/servicecache/service_cache.go
```diff
@@ -0,0 +1,174 @@
...
+
+	pb "github.com/cilium/hubble/api/v1/flow"
+
+	"github.com/cilium/cilium/api/v1/models"
+)
+
+// ServiceCache is a cache of existing services.
+type ServiceCache struct {
+	mu    sync.RWMutex
+	cache map[string]*entry
```
gandro: nit: Have you considered using two maps, e.g.:

```go
// index by numerical id
servicesByID map[int64]*entry
// index by frontend "<ip>:<port>"
servicesByAddr map[string]*entry
```

It would save you at the `genIDKey` function.

cmd/serve/serve.go
```diff
@@ -426,10 +432,12 @@ func consumeMonitorEvents(s *server.ObserverServer, conn net.Conn, version liste
 				monitorAPI.AgentNotifyEndpointRegenerateSuccess,
 				monitorAPI.AgentNotifyEndpointDeleted:
 				endpointEvents <- an
-			case monitorAPI.AgentNotifyIPCacheUpserted:
-				ipCacheEvents <- an
-			case monitorAPI.AgentNotifyIPCacheDeleted:
+			case monitorAPI.AgentNotifyIPCacheUpserted,
```
gandro: Good catch!

pkg/servicecache/service_cache.go
Rolinh: My initial implementation was using 2 maps but it turns out this was turning the process of writing tests into an even more painful experience as you always have to assert equality of two maps instead of just one within the tests. For the negligible overhead that the `genIDKey` function brings, I thought that it was a good trade-off to use only one map instead.

pkg/parser/getters/getters.go
Rolinh: Not really. I thought I would need to use it at some point but it turns out than it wasn't necessary and I just left the code there as it was already implemented/tested.
I think that in a way, it makes it clear that you can retrieve a service in the cache either by address or by ID but if you prefer, I have no strong feelings about removing it.

pkg/servicecache/service_cache.go
gandro: Sounds good to me!

pkg/parser/getters/getters.go
gandro: Sounds good, let's leave it in. :+1:

pkg/parser/seven/parser.go
Rolinh: The reason I picked `int` is because Go's [`net`](https://golang.org/pkg/net/) package uses `int` for port. However, I can change it to `uint16`. Note though that I have seen `uint32` being used for the port in the codebase as well. On a 64-bit platform, `int` defaults to `int64` and thus `uint32` can fit into. However, if I change from `int` to `uint16` as you suggest, we can in theory run into problems with `uint32` ports.

gandro: Ah! The `uint32` is due to  protobuf not having any 16bit integer types - I ideally we would have used `uint16` there as well, but we opted for uint32 instead: https://developers.google.com/protocol-buffers/docs/proto#scalar

 I leave it up to you if you want to change it. The Cilium service implementation itself is uses uint16 quite extensively, so using a `uint16` is basically guaranteed to be sufficient for the foreseeable future.

pkg/server/service.go
Rolinh: Nice catch!

pkg/parser/seven/parser.go
Rolinh: I wasn't aware of this protobuf limitation and it looks like `uint16` is used consistently for port outside of code interacting with protobuf. I'll make the change to `uint16` then.

michi-covalent:
lgtm!

pkg/parser/seven/parser.go
gandro: Awesome, thanks!!

pkg/servicecache/service_cache.go
```diff
@@ -0,0 +1,174 @@
...
+	ID           int64
+	Name         string
+	Namespace    string
+	FrontendIP   net.IP
+	FrontendPort uint16
+}
+
+// GetServiceByAddr retrieves a service from the cache given its frontend IP
+// and port. If the service was found in the cache, ok is true.
+func (svcc *ServiceCache) GetServiceByAddr(ip net.IP, port uint16) (service pb.Service, ok bool) {
```
glibsm: style-wise, these are usually just one char

pkg/servicecache/service_cache.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// New creates a new empty ServiceCache.
+func New() *ServiceCache {
+	return &ServiceCache{
+		cache: map[string]*entry{},
+	}
+}
+
+type entry struct {
```
glibsm: why does this un-exported struct have a bunch of exported members? usually that is only a pattern for (un)marshaling data from other formats (reflect requires public vars)

PR #64, opened 2020-01-23
serve: Add a "local" server

michi-covalent:
LocalObserverServer. It ignores all the state change events and does not
connect to Cilium's monitor socket. It's meant to be started embeeded in
Cilium process so that:

1. the parser can access Cilium state directly without making a copy, and
2. the server can receive monitor payloads by registering a listener.

The observer server interface adds a new method HandleMonitorSocket() that
encapsulates the logic for connecting to the monitor socket and handling
monitor events instead of exposing the logic in cmd/serve.go.

### DIFF
--- a/pkg/server/local_observer.go
+++ b/pkg/server/local_observer.go
@@ -1,0 +1,118 @@
+// Copyright 2020 Authors of Cilium
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package server
+
+import (
+	"context"
+
+	"github.com/cilium/cilium/pkg/math"
+	pb "github.com/cilium/hubble/api/v1/flow"
+	"github.com/cilium/hubble/api/v1/observer"
+	"github.com/cilium/hubble/pkg/container"
+	"github.com/cilium/hubble/pkg/parser"
+	"go.uber.org/zap"
+)
+
+// LocalObserverServer is an implementation of the server.Observer interface
+// that's meant to be run embedded inside the Cilium process. It ignores all
+// the state change events since the state is available locally.
+type LocalObserverServer struct {
+	// ring buffer that contains the references of all flows
+	ring *container.Ring
+
+	// events is the channel used by the writer(s) to send the flow data
+	// into the observer server.
+	events chan *pb.Payload
+
+	// stopped is mostly used in unit tests to signalize when the events
+	// channel is empty, once it's closed.
+	stopped chan struct{}
+
+	log *zap.Logger
+
+	// channel to receive events from observer server.
+	eventschan chan *observer.GetFlowsResponse
+
+	// payloadParser decodes pb.Payload into pb.Flow
+	payloadParser *parser.Parser
+}
+
+// NewLocalServer returns a new local observer server.
+func NewLocalServer(
+	payloadParser *parser.Parser,
+	maxFlows int,
+	logger *zap.Logger,
+) *LocalObserverServer {
+	return &LocalObserverServer{
+		log:  logger,
+		ring: container.NewRing(maxFlows),
+		// have a channel with 1% of the max flows that we can receive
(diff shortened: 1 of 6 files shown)


pkg/server/local_observer.go
```diff
@@ -0,0 +1,118 @@
...
+	// channel is empty, once it's closed.
+	stopped chan struct{}
+
+	log *zap.Logger
+
+	// channel to receive events from observer server.
+	eventschan chan *observer.GetFlowsResponse
+
+	// payloadParser decodes pb.Payload into pb.Flow
+	payloadParser *parser.Parser
```
glibsm: Curious: why does local server need a payload parser?

pkg/server/local_observer.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// NewLocalServer returns a new local observer server.
+func NewLocalServer(
+	payloadParser *parser.Parser,
+	maxFlows int,
+) *LocalObserverServer {
+	return &LocalObserverServer{
+		log:  logger.GetLogger(),
```
glibsm: accepting log as a param (injecting) is much better than a global function

pkg/server/local_observer.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/cilium/hubble/pkg/container"
+	"github.com/cilium/hubble/pkg/logger"
+	"github.com/cilium/hubble/pkg/parser"
+	"go.uber.org/zap"
+)
+
+// LocalObserverServer is an implementation of the server.Observer interface
+// that's meant to be run embedded inside the Cilium process. It ignores all
+// the state change events since the state is available locally.
+type LocalObserverServer struct {
```
glibsm: there is still a lot of duplication between the two (implementing a lot of the same getters), but this is an improvement over the previous PR i think

pkg/server/local_observer_test.go
```diff
@@ -0,0 +1,103 @@
...
+	assert.NotNil(t, s.GetStopped())
+	assert.NotNil(t, s.GetPayloadParser())
+	assert.NotNil(t, s.GetRingBuffer())
+	assert.NotNil(t, s.GetLogger())
+	assert.NotNil(t, s.GetEventsChannel())
+}
+
+func TestLocalObserverServer_ServerStatus(t *testing.T) {
+	pp, err := parser.New(
+		&testutils.NoopEndpointGetter,
```
glibsm: Not sure why these are extracted into testutils, seems like they are related to this package only no?

pkg/server/observer.go
```diff
@@ -15,14 +15,27 @@
...
 	"context"
+	"encoding/gob"
 	"fmt"
+	"io"
+	"log"
 	"net"
 	"strings"
 	"time"
 
+	"github.com/cilium/cilium/pkg/defaults"
```
glibsm: 🤔

pkg/server/local_observer.go
michi-covalent: monitor listener receives payloads like this: https://github.com/cilium/cilium/commit/92c6cf120a44699e98b0bae233e8640f20d1065d#diff-eb79fc32faeb56e625dc10a96f3af5a1R35 so we still need to apply the same parsing logic.

pkg/server/local_observer_test.go
michi-covalent: they are used outside of the server package too, pkg/parser/seven/parser_test.go for example.

pkg/server/observer.go
michi-covalent: fixed!

pkg/server/local_observer.go
michi-covalent: yeah i almost think that ObserverServer can have LocalObserverServer as a field, but maybe as a follow-up pr...

michi-covalent: something like this: https://github.com/cilium/hubble/commit/30c9ef6b4242025d85f291fc792707d88370b0dc

pkg/server/local_observer.go
michi-covalent: fixed

pkg/server/local_observer_test.go
```diff
@@ -0,0 +1,104 @@
+// Copyright 2019 Authors of Cilium
```
gandro: nit: Copyright 2020

pkg/server/local_observer.go
```diff
@@ -0,0 +1,118 @@
+// Copyright 2019 Authors of Cilium
```
gandro: nit: Copyright 2020

pkg/server/local_observer_test.go
michi-covalent: done

pkg/server/local_observer.go
michi-covalent: done

Issue #50, opened 2020-01-14
migrate to Go modules


Rolinh:
Cilium already migrated to Go modules (see https://github.com/cilium/cilium/pull/8719). Hubble should also migrate from dep to go mod for dependency management.

Rolinh:
Currently, one cannot "go get" cilium which prevents migrating to go mod. I need to investigate this further.
```
$ go version
go version go1.13.6 linux/amd64
$ go get -u github.com/cilium/cilium
go: github.com/cilium/cilium@v1.7.0-rc2 requires
        github.com/optiopay/kafka@v0.0.0-20180809090225-01ce283b732b: invalid version: unknown revision 01ce283b732b
```

bobcob7:
I cannot seem to 'go get' this.
```
$ go version
go version go1.15.4 linux/amd64
$ go get -u github.com/cilium/cilium
go: github.com/cilium/cilium upgrade => v1.8.5
go get: github.com/cilium/cilium@v1.8.5 requires
        github.com/optiopay/kafka@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
```

Perhaps this can be resolved by changing the replace entry to [v1.5.0](https://github.com/optiopay/kafka/tree/v1.5.0) in go.mod.

bobcob7:
In case anyone else runs into this, I found that my issue was caused by the lack of git in my path.

PR #67, opened 2020-01-28
vendor: migrate to go-mod

Rolinh:
Note that replace directives set in `go.mod` files from dependencies (eg: `github.com/cilium/cilium`) are not taken into account by `go mod` (see https://github.com/golang/go/issues/30354#issuecomment-466479708).

Cilium's `go.mod` replace directives for k8s thus have to be replicated in hubble's `go.mod` file.

Closes: #50


michi-covalent:
alright i'm merging this

PR #69, opened 2020-01-28
server: Refactor LocalObserverServer and ObserverServer

michi-covalent:
Use LocalObserverServer in ObserverServer instead of duplicating the
same code, and move the gRPC server code to LocalObserverServer. Define
an interface GRPCServer so that we can use different implementation for
unit tests etc..

With this change, there is a clear split between LocalObserverServer
and ObserverServer. LocalObserverServer is responsible for maintaining
the ring buffer and implementing gRPC API, while ObserverServer is
responsible for consuming monitor events and syncing Cilium state.

### DIFF
--- a/cmd/serve/serve.go
+++ b/cmd/serve/serve.go
@@ -31,6 +31,7 @@ import (
 	"github.com/cilium/hubble/pkg/cilium/client"
 	"github.com/cilium/hubble/pkg/fqdncache"
 	"github.com/cilium/hubble/pkg/ipcache"
+	"github.com/cilium/hubble/pkg/logger"
 	"github.com/cilium/hubble/pkg/metrics"
 	metricsAPI "github.com/cilium/hubble/pkg/metrics/api"
 	"github.com/cilium/hubble/pkg/parser"
@@ -97,10 +98,16 @@ func New(log *zap.Logger) *cobra.Command {
 				serviceCache,
 				payloadParser,
 				int(maxFlows),
+				logger.GetLogger(),
 			)
-			err = Serve(log, listenClientUrls, s)
+			s.Start()
+			err = Serve(log, listenClientUrls, s.GetGRPCServer())
 			if err != nil {
 				log.Fatal("", zap.Error(err))
+			}
+			fmt.Printf("Press Ctrl-C to quit\n")
+			if err := s.HandleMonitorSocket(nodeName); err != nil {
+				log.Fatal("HandleMonitorSocket failed", zap.Error(err))
 			}
 		},
 	}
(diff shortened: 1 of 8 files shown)


pkg/server/endpoint_test.go
```diff
@@ -152,6 +153,16 @@ func (f *fakeEndpointsHandler) GarbageCollect() {
...
 }
 
+func getLocalObserverServer() *LocalObserverServer {
+	pp, _ := parser.New(
+		&testutils.NoopEndpointGetter,
+		&testutils.NoopIdentityGetter,
+		&testutils.NoopDNSGetter,
+		&testutils.NoopIPGetter,
+		&testutils.NoopServiceGetter)
+	return NewLocalServer(pp, 10, zap.L())
```
glibsm: zap has a nop logger as well

cmd/serve/serve.go
```diff
@@ -229,14 +236,13 @@ func setupListeners(listenClientUrls []string) (listeners map[string]net.Listene
 
 // Serve starts the GRPC server on the provided socketPath. If the port is non-zero, it listens
 // to the TCP port instead of the unix domain socket.
-func Serve(log *zap.Logger, listenClientUrls []string, s server.Observer) error {
+func Serve(log *zap.Logger, listenClientUrls []string, s *server.LocalObserverServer) error {
```
glibsm: This swap seems like it's headed in the wrong direction. We used to accept an interface, and now we're taking a pointer to a concrete struct.

pkg/server/local_observer.go
```diff
@@ -92,27 +114,202 @@ func (s *LocalObserverServer) GetStopped() chan struct{} {
...
+
+		for e := range ch {
+			if req.Number != 0 && !req.Follow {
+				i++
+				if i >= req.Number {
+					// stop the channel buffer because we have reached
+					// the number of requested flows.
+					stop()
+					if i > req.Number {
+						// We will 'continue' since 'ch' might have flows and we
```
glibsm: I don't see a `continue` here, I see a return 🤔

pkg/server/endpoint_test.go
michi-covalent: updated

cmd/serve/serve.go
michi-covalent: it used to take an interface because LocalObserverServer and ObserverServer implemented the same interface with the duplicated code.

now there is only one implementation of local observer server, so i didn't bother to keep the interface. i can define an interface for the local observer server if you think that might be useful. let me know.

pkg/server/local_observer.go
michi-covalent: 🤔 it came from https://github.com/cilium/hubble/pull/69/commits/fceda102a528ee3c3cae18b0da5dfc5921a4f7f3#diff-69945783af04950df9270c268c24c6a1L439 we can fix it in a follow-up commit. i prefer to keep the original code for a refactor commit.

michi-covalent:
ping!

michi-covalent:
@Rolinh yeah i was thinking the same thing. i'll come up with better names for them.

michi-covalent:
@Rolinh i changed the interface name from `LocalObserver` to `GRPCServer`, and also the field name in `ObserverServer` from `localServer` to `grpcServer`.

i'm deferring renaming of actual structs (`ObserverServer`/`LocalObserverServer`) to a follow-up PR to avoid introducing too many changes in this PR.

PR #72, opened 2020-01-31
server: refactor getflows

### LINKED ISSUE #10: Consider reversing output when using `--since`
Hubble currently output flows with the latest flow first which requires to read request and response interactions from bottom to top. This seems unnatural.

We should consider reversing the order in which flows are being printed.

### LINKED ISSUE #68: apply filter before applying `--last`
The current logic gets the `--last` number of elements from the ring buffer, then applies the filter which usually results in less than the requested amount of flows or even none.

Rolinh:
This PR introduces `RingReader` and refactors the get flows logic.

Fixes #10, Fixes #68

### DIFF
--- a/pkg/container/ring.go
+++ b/pkg/container/ring.go
@@ -1,4 +1,4 @@
-// Copyright 2019 Authors of Hubble
+// Copyright 2019-2020 Authors of Hubble
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -12,27 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package container implements a ring buffer that can be accessible by multiple
-// readers.
-//
-// To iterate over a PayloadRing (where t is the PayloadRing):
-// 	for lastWritePos := t.LastWriteParallel(); ; lastWritePos-- {
-// 	    value, stop := s.Read(lastWriteIdx)
-//      if stop {
-//          break
-//      }
-//      // do something with value
-// 	}
-//
-// Write is *not* being executed in parallel you can use
-// 	for lastWritePos := t.LastWrite(); ; lastWritePos-- {
-// 	    value, stop := s.Read(lastWriteIdx)
-//      if stop {
-//          break
-//      }
-//      // do something with value
-// 	}
-//
+// Package container implements a ring buffer that can be accessed by multiple
+// RingReader readers.
 package container
 
 import (
@@ -139,9 +120,9 @@ func (r *Ring) LastWrite() uint64 {
 	return atomic.LoadUint64(&r.write) - 1
 }
 
-// Read reads the *v1.Event from the given read position. Returns false if
+// read reads the *v1.Event from the given read position. Returns false if
 // that position is no longer available to be read, returns true otherwise.
-func (r *Ring) Read(read uint64) (*v1.Event, bool) {
+func (r *Ring) read(read uint64) (*v1.Event, bool) {
 	readIdx := read & r.mask
 	flow := r.data[readIdx]
 
@@ -173,9 +154,9 @@ func (r *Ring) Read(read uint64) (*v1.Event, bool) {
 	return nil, false
 }
 
-// ReadFrom continues to read from the given position until the stop channel
+// readFrom continues to read from the given position until the stop channel
 // is closed.
-func (r *Ring) ReadFrom(stop <-chan struct{}, read uint64) <-chan *v1.Event {
+func (r *Ring) readFrom(stop <-chan struct{}, read uint64) <-chan *v1.Event {
 	// TODO should we create the channel or the caller?
 	const returnedBufferChLen = 1000
 	ch := make(chan *v1.Event, returnedBufferChLen)
--- a/pkg/container/ring_reader.go
+++ b/pkg/container/ring_reader.go
@@ -1,0 +1,68 @@
+// Copyright 2020 Authors of Hubble
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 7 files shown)


pkg/container/ring_reader_test.go
```diff
@@ -0,0 +1,177 @@
...
+				ctx, cancel := context.WithTimeout(context.Background(), 1*time.Millisecond)
+				got = append(got, reader.Previous(ctx))
+				select {
+				case <-ctx.Done():
+					timedOut = true
+				default:
+				}
+				cancel()
+			}
+			if !reflect.DeepEqual(got, tt.want) {
```
glibsm: `assert.Equal(t, tt.want, got)`

pkg/server/observer.go
```diff
@@ -289,124 +289,166 @@ func (s *ObserverServer) GetFlows(
...
+func newFlowsReader(r *container.RingReader, req *observer.GetFlowsRequest, log *zap.Logger) (*flowsReader, error) {
+	whitelist, err := filters.BuildFilterList(req.Whitelist)
+	if err != nil {
+		return nil, err
+	}
+	blacklist, err := filters.BuildFilterList(req.Blacklist)
+	if err != nil {
+		return nil, err
+	}
+	log.Debug("filters", zap.String("req", fmt.Sprintf("%+v", req)))
```
michi-covalent: nit: maybe combine into a single log

```
log.Debug("processing getflows request", 
    zap.Any("req", req),
    zap.Any("whitelist", whitelist),
    zap.Any("blacklist", blacklist"))
```

pkg/server/observer.go
```diff
@@ -289,124 +289,166 @@ func (s *ObserverServer) GetFlows(
...
 ) (err error) {
-	reply, err := getFlows(server.Context(), obs.GetLogger(), obs.GetRingBuffer(), req)
+	start := time.Now()
+	log := obs.GetLogger()
+	ring := obs.GetRingBuffer()
+
+	i := uint64(0)
+	defer func() {
+		size := ring.Cap()
+		took := time.Now().Sub(start)
```
glibsm: can inline these into the log, no need for new vars

pkg/server/observer.go
```diff
@@ -289,124 +289,166 @@ func (s *ObserverServer) GetFlows(
...
 	}
 }
 
-func getUntil(req *observer.GetFlowsRequest, defaultTime *types.Timestamp) (time.Time, error) {
-	until := req.GetUntil()
-	if until == nil {
-		until = defaultTime
-	}
-	return types.TimestampFromProto(until)
+type flowsReader struct {
```
glibsm: nesting two "readers" probably warrants a comment who abstracts what

pkg/server/observer.go
```diff
@@ -416,52 +458,39 @@ func getFlows(
...
-			select {
-			case reply <- flow:
-				// We have sent all expected flows so we can return already
-				if req.Number != 0 && i >= req.Number {
-					return
-				}
-			case <-ctx.Done():
-				return
+			if ts.Before(start) {
+				idx++ // we went backward 1 too far
```
glibsm: This is quite prone to races (since writing happens very fast, concurrently).

If you got back to the oldest entry and will now ++ it to return, it's very likely it's going to be ovrewritten.

I'm not sure that there is a good way to resolve it without read locking elements (which is probably not necessary here). Perhaps we should just buffer more than one element on the end. Maybe we do a % cut-off, like 1,2,3,5% at the end of the reader we assume is for "staging/rewriting" area.

What do you think?

michi-covalent:
lgtm! let's give others some time to review as well. i'll merge it in monday morning if there is no objection.

pkg/server/observer_test.go
```diff
@@ -348,8 +348,8 @@ func TestObserverServer_GetFlowsBetween(t *testing.T) {
 		t.Errorf("GetFlowsBetween error = %v, wantErr %v", err, nil)
 	}
 
-	for i := 0; i < 4; i++ {
-		assert.Equal(t, int64(6-i), got[i].GetFlow().Time.Seconds)
+	for i := 0; i < 6; i++ {
+		assert.Equal(t, int64(i+2), got[i].GetFlow().Time.Seconds)
```
glibsm: This change in the test doesn't scream at me "the order is now reversed"

pkg/container/ring.go
```diff
@@ -16,22 +16,22 @@
...
-// 	for lastWritePos := t.LastWrite(); ; lastWritePos-- {
-// 	    value, stop := s.Read(lastWriteIdx)
+//	for lastWritePos := t.LastWrite(); ; lastWritePos-- {
+//	    value, stop := s.Read(lastWriteIdx)
 //      if stop {
 //          break
 //      }
 //      // do something with value
-// 	}
+//	}
```
glibsm: Can probably update docs to use the reader. In fact, we may want to hide the ring to be private and only expose the reader

pkg/container/ring_reader.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"context"
+	"sync"
+
+	"github.com/cilium/hubble/pkg/api/v1"
+)
+
+// RingReader is a reader for a Ring container.
+type RingReader struct {
```
glibsm: hmmmm... looks familiar 🤔

pkg/server/observer.go
Rolinh: This code is as it was but I can sure inline this.

pkg/server/observer.go
Rolinh: Good point, I'll add  doc to `flowsReader`.

pkg/server/observer_test.go
Rolinh: Does it need to?

pkg/server/observer.go
Rolinh: `newFlowsReader` just creates a new `flowsReader` and doesn't actually process anything. Nonetheless, you have a point: I haven't updated the debug log to reflect that.

pkg/server/observer.go
Rolinh: I think that this whole code basically is but don't really see it as a problem in this context in the sense that it's no worse that it was.
For sure, I think that we don't want to create lock contention on the ring buffer so your idea of a staging area is probably worth exploring.

pkg/server/observer_test.go
glibsm: Ideally tests are easy to read and reason about, yes. This is more of a comment on the test setup though

PR #74, opened 2020-02-04
endpoint: Add a function to lookup by container ID

michi-covalent:
Also add .idea directory to gitignore.

### DIFF
--- a/pkg/api/v1/endpoint.go
+++ b/pkg/api/v1/endpoint.go
@@ -196,3 +196,20 @@ func (es *Endpoints) GetEndpoint(ip net.IP) (endpoint *Endpoint, ok bool) {
 	}
 	return
 }
+
+// GetEndpointByContainerID returns the endpoint that has the given container ID.
+func (es *Endpoints) GetEndpointByContainerID(id string) (*Endpoint, bool) {
+	es.mutex.RLock()
+	defer es.mutex.RUnlock()
+	for _, ep := range es.eps {
+		if ep.Deleted != nil {
+			continue
+		}
+		for _, containerID := range ep.ContainerIDs {
+			if id == containerID {
+				return ep, true
+			}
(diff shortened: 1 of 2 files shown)


pkg/api/v1/endpoint.go
```diff
@@ -196,3 +196,20 @@ func (es *Endpoints) GetEndpoint(ip net.IP) (endpoint *Endpoint, ok bool) {
...
+func (es *Endpoints) GetEndpointByContainerID(id string) (*Endpoint, bool) {
+	es.mutex.RLock()
+	defer es.mutex.RUnlock()
+	for _, ep := range es.eps {
+		if ep.Deleted != nil {
+			continue
+		}
+		for _, containerID := range ep.ContainerIDs {
+			if id == containerID {
+				return ep, true
```
Rolinh: `es.eps` is a slice of pointer to `Endpoint`s. The caller of that method can thus modify the `Endpoint` and it's probably not something that should be allowed.

michi-covalent: fair point. we do the same thing in GetEndpoint() as well. should we change it to return a copy instead?

Rolinh: I think it should eventually be tackled as part of #76 but for the time being, returning a copy and adding a `FIXME` note seems reasonable to me.

michi-covalent: added another commit to make a copy before returning

.gitignore
```diff
@@ -1,2 +1,3 @@
 /hubble
 install/kubernetes/hubble.yaml
+.idea/
```
glibsm: This doesn't hurt to add, but this belongs in your global .gitignore, really.

michi-covalent: fair enough. reverted.

PR #98, opened 2020-02-11
docs: update main architecture diagram

genbit:
Design of main diagram was a bit updated by @summer-street
hubble_arch

Signed-off-by: Sergey Generalov <sergey@genbit.ru>


glibsm:
`./jq` feels extraneous to the diagram

genbit:
> `./jq` feels extraneous to the diagram

yeah, it was there in the old diagram. This is mainly a design improvement

genbit:
> I still think we should drop `./jq`

let's do it as separate PR or something.

PR #102, opened 2020-02-13
server: ensure to never block if `--follow` is not provided

Rolinh:


pkg/server/local_observer.go
```diff
@@ -303,7 +303,14 @@ func (r *flowsReader) Next(ctx context.Context) (*pb.Flow, error) {
 		if r.maxFlows > 0 && !r.follow && (r.flowsCount >= r.maxFlows) {
 			return nil, io.EOF
 		}
-		e := r.ringReader.Next(ctx)
+		readCtx, cancelReadCtx := context.WithTimeout(ctx, 50*time.Millisecond)
```
gandro: Blocking for 50ms seems a bit unfortunate. Would it be possible to have a `ringReader.NextNonBlocking()` which would look something like this:
```go
func (r *RingReader) NextNonBlocking() *v1.Event {
    e, ok := r.ring.read(r.idx)
    if ok {
        r.idx++
        return e
    }
    return nil
}
```

Rolinh:
@glibsm @gandro @michi-covalent I implemented your feedback, PTAL.

pkg/container/ring_reader.go
```diff
@@ -39,20 +39,29 @@ func NewRingReader(ring *Ring, start uint64) *RingReader {
 }
 
 // Previous reads the event at the current position and decrement the read
-// position by one.
-func (r *RingReader) Previous(ctx context.Context) *v1.Event {
+// position by one. If no event was read, ok is false.
+func (r *RingReader) Previous() (event *v1.Event, ok bool) {
```
glibsm: isn't nil event the same as ok false here? or what's the difference now between nil and false

Rolinh: Unfortunately no. The ring may return `nil/true` on `read()` when the buffer is not full and without the `ok` there's no way to tell if one should keep reading or no.

glibsm: I guess what I'm trying to say is we can make this function promise to never return `nil, true` as if everything is fine. Then consumers don't ever have to check for both ok and nil, right?

Rolinh: Good point, I'll implement.

Rolinh:
@glibsm PTAL

PR #105, opened 2020-02-14
Allow for setting the priority class and resource requests/limits.

bmcstdio:
This PR adds the ability to set (in the case of the Hubble UI) and override (in the case of Hubble itself) the priority class being used, as well as the ability to set resource requests and limits.

### DIFF
--- a/install/kubernetes/hubble/Chart.yaml
+++ b/install/kubernetes/hubble/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: hubble
-version: 0.0.1
+version: 0.0.2
 appVersion: 0.0.1
 tillerVersion: ">=2.7.2"
 description: Helm chart for the Hubble server
--- a/install/kubernetes/hubble/values.yaml
+++ b/install/kubernetes/hubble/values.yaml
@@ -49,6 +49,8 @@ ui:
     pullPolicy: Always
   clusterDomain: cluster.local
   replicas: 1
+  resources: {}
+  priorityClassName: ""
 
 ingress:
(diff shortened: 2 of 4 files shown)


install/kubernetes/hubble/values.yaml
```diff
@@ -49,6 +49,8 @@ ui:
     pullPolicy: Always
   clusterDomain: cluster.local
   replicas: 1
+  resources: {}
```
michi-covalent: could you add a comment here with a sample value?

install/kubernetes/hubble/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: hubble
-version: 0.0.1
+version: 0.0.2
```
glibsm: We haven't been very diligent about actually nudging this in the past, thanks for pointing this out.

install/kubernetes/hubble/values.yaml
```diff
@@ -62,3 +64,6 @@ ingress:
   #  - secretName: chart-example-tls
   #    hosts:
   #      - chart-example.local
+
+resources: {}
+priorityClassName: system-node-critical
```
Rolinh: Would you mind adding a comment here? Something like the following:
> The priority class `system-node-critical`  marks  add-on pods as critical to the node itself. This priority class is only valid under the `kube-system` namespace.

bmcstdio:
@michi-covalent @glibsm @Rolinh thank you for your reviews and suggestions. I've addressed all your comments.

PR #116, opened 2020-02-18
serve: Export EnableMetrics function

michi-covalent:
- Export EnableMetrics so that Cilium can use it.
- Take metricsServer as an argument.


michi-covalent:
cc @tgraf

PR #117, opened 2020-02-19
Add options to the flow metrics

ap4y:
Flow plugins doesn't expose source and destination prometheus
labels. This patch will add support for the context options to support
additional prometheus labels.

Signed-off-by: Arthur Evstifeev <aevstifeev@gitlab.com>

### DIFF
--- a/pkg/metrics/flow/handler.go
+++ b/pkg/metrics/flow/handler.go
@@ -15,7 +15,7 @@
 package flow
 
 import (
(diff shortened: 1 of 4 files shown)


michi-covalent:
thank you for the pr @ap4y ! the pr looks good to me. @tgraf could you also take a look? i guess we simply forgot about including labels for flows metric?

michi-covalent:
ok i'm merging this. thanks again @ap4y !

Issue #97, opened 2020-02-11
Allow embedding Hubble into Cilium


tgraf:

michi-covalent:
PoC branch: https://github.com/cilium/cilium/commits/pr/michi/hubble2

need to clean up before opening a PR, but it would be nice to get some feedback to make sure it looks reasonable, especially this commit: https://github.com/cilium/cilium/commit/259dc6bfa860f48345d9f36b45061b1ece0d1ae7

cc @tgraf @glibsm @gandro @Rolinh

tgraf:
This has been merged into Cilium via https://github.com/cilium/cilium/pull/10238

PR #120, opened 2020-02-19
logger: Avoid calling logger.GetLogger()

michi-covalent:
Typically logger gets configured externally (e.g. by Cilium) in embeded
mode. Hubble needs to use this logger across all the components instead
of using logger.GetLogger().

Ref #97


michi-covalent:
let me think about this a bit more. closing it for now

PR #132, opened 2020-02-24
cmd/observe: deprecate --compact, --dict, --json flags, add --output

### LINKED ISSUE #121: --follow doesn't work with --dict
`hubble observe --follow --dict` should follow in dictionary mode, yet it incorrectly defaults to compact.

`--follow --json` works correctly and follows in json

Rolinh:
This makes it more easy to find out about all available out formatting options. Furthermore, when old formatting flags will be removed, providing flags that are incompatible, such as `--json --compact`, will not be possible anymore. For now however, just deprecate the old formatting flags in order to not to break compatibility.

This also fixes a bug where `--follow --dict` was incorrectly defaulting to the compact output.

Fixes #121

### DIFF
--- a/cmd/observe/observe.go
+++ b/cmd/observe/observe.go
@@ -45,6 +45,7 @@ var (
 	jsonOutput            bool
 	compactOutput         bool
 	dictOutput            bool
+	output                string
 	follow                bool
 	ignoreStderr          bool
 	enableIPTranslation   bool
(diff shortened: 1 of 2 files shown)


cmd/observe/observe.go
```diff
@@ -210,14 +208,9 @@ programs attached to endpoints and devices. This includes:
...
-		&jsonOutput, "json", "j", false, "Use json output",
-	)
-	observerCmd.Flags().BoolVar(
-		&compactOutput, "compact", false, "Use compact output. Automatically used with --follow, unless --json is specified.",
-	)
-	observerCmd.Flags().BoolVar(
-		&dictOutput, "dict", false, "Use dictionary output. Each flow will be shown as KEY:VALUE pairs.",
+	observerCmd.Flags().StringVarP(
+		&output, "output", "o", "compact",
+		"Specify the output format, one of:\n -compact:\tcompact output\n -dict:\teach flow is shown as KEY:VALUE pair\n -json:\tJSON encoding\n",
```
gandro: Note that there is also a fourth mode, which is actually the current default: The non-compact table mode (selected if `var opts []hubprinter.Option` contains no other mode), which I think should be stay default if `--follow` is absent. We only switch to compact mode when `--follow` is provided.

pkg/printer/options.go
```diff
@@ -70,6 +70,13 @@ func Dict() Option {
 	}
 }
 
+// Tab prints flows in even tab-aligned columns.
+func Tab() Option {
```
glibsm: tab is the `iota` for output enum, not sure this is really needed

PR #137, opened 2020-02-25
container/ring_reader: never return nil events on NextFollow()

### LINKED ISSUE #134: observe --follow terminates when buffer is empty
this might be related to #131, but i'm seeing a slightly different issue where `observe --follow` terminates when ring buffer is *empty*. i looked into it a bit, basically what happens is:

- flow is nil here: https://github.com/cilium/hubble/blob/00dc35a7f50dedea536c2b987b9fce1f8a079556/pkg/container/ring.go#L180
- NextFollow returns nil event: https://github.com/cilium/hubble/blob/00dc35a7f50dedea536c2b987b9fce1f8a079556/pkg/container/ring_reader.go#L75
- local observer returns io.EOF: https://github.com/cilium/hubble/blob/00dc35a7f50dedea536c2b987b9fce1f8a079556/pkg/server/local_observer.go#L311
- getFlows terminates: https://github.com/cilium/hubble/blob/00dc35a7f50dedea536c2b987b9fce1f8a079556/pkg/server/local_observer.go#L201

Rolinh:
By contract, RingReader should only return <nil> when no event can be read. In the case of NextFollow(), one should block until an event is read or the context is cancelled.
However, it appears that under certain circumstances, typically when the ring buffer is empty, <nil> events may be sent to the channel by Ring.readFrom(). In such a case, one should ignore the <nil> event and keep reading/waiting until a non-nil event is being sent to the channel or the context is cancelled.

Fixes #134 

/cc @michi-covalent Can you please confirm that if fixes the issue you reported?

### DIFF
--- a/pkg/container/ring_reader.go
+++ b/pkg/container/ring_reader.go
@@ -69,11 +69,17 @@ func (r *RingReader) NextFollow(ctx context.Context) *v1.Event {
 	if r.c == nil {
 		r.c = r.ring.readFrom(r.stop, r.idx)
 	}
-	select {
-	case e := <-r.c:
-		r.idx++
-		return e
-	case <-ctx.Done():
-		return nil
+	// when the ring is not full, nil events may be sent to the channel
+	// one should keep reading in such a case
+	for {
+		select {
+		case e := <-r.c:
+			r.idx++
+			if e != nil {
+				return e
+			}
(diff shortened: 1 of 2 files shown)


michi-covalent:
thanks @Rolinh i'll test this today

PR #139, opened 2020-02-25
mod: Regenerate go.mod

michi-covalent:
We've accumlated a bunch of packages in go.mod that are no longer
relevant, but somehow `go mod tidy` doesn't remove these packages.


michi-covalent:
ok guys i'm going in

go.mod
```diff
@@ -3,32 +3,23 @@ module github.com/cilium/hubble
...
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.6.1
 	github.com/stretchr/testify v1.4.0
-	google.golang.org/genproto v0.0.0-20190801165951-fa694d86fc64 // indirect
 	google.golang.org/grpc v1.23.1
 	k8s.io/apimachinery v0.17.3
 )
 
-replace (
-	github.com/miekg/dns => github.com/cilium/dns v1.1.4-0.20190417235132-8e25ec9a0ff3
```
Rolinh: I think this could bite us at some point. This is a dependency of cilium and `go get` basically does not read replace directives outside the main module. In effect, this means that this dependency for cilium is different that the one that cilium actually expects. See the diff in `vendor/modules.txt`:
```
- # github.com/miekg/dns v1.1.22 => github.com/cilium/dns v1.1.4-0.20190417235132-8e25ec9a0ff3
+ # github.com/miekg/dns v1.0.14
```
I think that the only real fix for this, if these forks are "permanent", is to directly refer to the forks within cilium's codebase rather than using a replace directive (which is a feature intended for dev purposes; not the way it is used in cilium now (see [FAQ](https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive)).

Issue #82, opened 2020-02-06
container (ring): data race while running tests


Rolinh:
Hit a data race while running the tests with the `-race` flag:
```
$ go test -timeout=30s -cover -mod=vendor -race $(go list ./...)
(...)
==================
WARNING: DATA RACE
Read at 0x00c000350808 by goroutine 92:
  github.com/cilium/hubble/pkg/container.(*Ring).readFrom.func1()
      /home/vagrant/go/src/github.com/cilium/hubble/pkg/container/ring.go:180 +0x1c0

Previous write at 0x00c000350808 by goroutine 99:
  github.com/cilium/hubble/pkg/container.(*Ring).Write()
      /home/vagrant/go/src/github.com/cilium/hubble/pkg/container/ring.go:102 +0xd8
  github.com/cilium/hubble/pkg/server.(*LocalObserverServer).Start()
      /home/vagrant/go/src/github.com/cilium/hubble/pkg/server/local_observer.go:109 +0x6ad

Goroutine 92 (running) created at:
  github.com/cilium/hubble/pkg/container.(*Ring).readFrom()
      /home/vagrant/go/src/github.com/cilium/hubble/pkg/container/ring.go:163 +0xbc
  github.com/cilium/hubble/pkg/container.(*RingReader).Next()
      /home/vagrant/go/src/github.com/cilium/hubble/pkg/container/ring_reader.go:57 +0x281
  github.com/cilium/hubble/pkg/server.(*flowsReader).Next()
      /home/vagrant/go/src/github.com/cilium/hubble/pkg/server/local_observer.go:306 +0x16e
  github.com/cilium/hubble/pkg/server.getFlows()
      /home/vagrant/go/src/github.com/cilium/hubble/pkg/server/local_observer.go:199 +0x3ee
  github.com/cilium/hubble/pkg/server.(*LocalObserverServer).GetFlows()
      /home/vagrant/go/src/github.com/cilium/hubble/pkg/server/local_observer.go:162 +0x7e
  github.com/cilium/hubble/pkg/server.TestObserverServer_GetLastNFlows_With_Follow.func3()
      /home/vagrant/go/src/github.com/cilium/hubble/pkg/server/observer_test.go:249 +0x8f

Goroutine 99 (running) created at:
  github.com/cilium/hubble/pkg/server.(*ObserverServer).Start()
      /home/vagrant/go/src/github.com/cilium/hubble/pkg/server/observer.go:122 +0xf8
==================
[...]

PR #142, opened 2020-02-25
reduce timeout for tests and make it configurable; run tests with -race flag

Rolinh:
This PR will fail until #82 is fixed.


gandro:
This should be unblocked now. Needs a rebase though.

PR #144, opened 2020-02-26
getter: Change EndpointGetter to return a subset of Endpoint

michi-covalent:
Define a struct EndpointInfo that contains endpoint ID, pod namespace,
pod name, and labels, and change EndpointGetter to return EndpointInfo
insted of the full Endpoint struct.

The main rationale for this change is to simplify the implementation of
the EndpointGetter in embedded mode. Since we don't need EndpointHandler
in embedded mode, EndpointGetter can return only the fields that are
used by the parser and LegacyPodGetter.

Ref #97

### DIFF
--- a/pkg/api/v1/endpoint.go
+++ b/pkg/api/v1/endpoint.go
@@ -235,6 +235,23 @@ func (es *Endpoints) GetEndpoint(ip net.IP) (endpoint *Endpoint, ok bool) {
 	return
 }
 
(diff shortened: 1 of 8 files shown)


glibsm:
I'm not sure I follow, why is hubble the source of information for cilium endpoints?

michi-covalent:
superseded by https://github.com/cilium/hubble/issues/150

PR #146, opened 2020-02-26
update Go to v1.14.0

Rolinh:
Note that `go mod vendor` had to be re-run to sync `vendor/modules.txt` as Go v1.14 reported inconsistent vendoring (missing `explicit` and `replaced` directives in `vendor/modules.txt`).

/cc @michi-covalent

PR #149, opened 2020-02-26
vendor: mark goleak as explicit

Rolinh:
Essentially, this means running `go mod vendor` with Go 1.14 (I should have re-sync #146 before merging). This fixes the current test failure on master.


glibsm:
is modules.txt meant to be touched manually? i thought it was all auto-generated

Rolinh:
It is auto-generated. However, #148 added the goleak dependency while still using Go 1.13.

glibsm:
Then this may continue flip flopping depending on who is running it :)

May be time to sneak something into the makefile that requires 1.14 for development

Rolinh:
I don't think it will because since #146 was merged, github actions uses Go 1.14.
So if someone were to open a PR that modifies a dependency in a way that makes Go 1.14 unhappy (eg: by still using Go 1.13), the check will fail then.

PR #151, opened 2020-02-26
make: remove -mod=vendor from go commands

Rolinh:
From the [Go 1.14 release notes](https://golang.org/doc/go1.14):

> When the main module contains a top-level vendor directory and its go.mod file specifies go 1.14 or higher, the go command now defaults to -mod=vendor for operations that accept that flag

Since the go.mod file now specifies go 1.14, it is no longer necessary to manually add the `-mod=vendor` flag to go commands.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ TARGET=hubble
 all: hubble
 
 hubble:
-	$(GO) build -mod=vendor -o $(TARGET)
+	$(GO) build -o $(TARGET)
 
 install:
 	groupadd -f hubble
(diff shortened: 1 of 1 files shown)


michi-covalent:
@glibsm i'm more on the side of enforcing the minimum version and cleaning up the stuff that are no longer necessary. @aanm also mentioned that there are some gofmt changes https://app.slack.com/client/T1MATJ4SZ/C2B917YHE in general i feel like trying to support multiple go versions for dev environment is a lost cause.

michi-covalent:
> It's currently not documented anywhere that we require 1.14

true. @Rolinh maybe document it somewhere or add a version check of some sort?

Rolinh:
@michi-covalent What about this? #153

Issue #150, opened 2020-02-26
Define an interface to access Endpoint fields


michi-covalent:
so that hubble and cilium (for embedded mode) can define the interface independently for endpoint getter. it needs access to the following fields:

- ID
- PodName
- PodNamespace
- Labels

PR #152, opened 2020-02-26
endpoint: Define an interface to access endpoint fields

michi-covalent:
This allows the embedded mode to use a different underlying endpoint
structure without having to convert to v1.Endpoint.

Ref #150


Rolinh:
@michi-covalent Doesn't this PR close #150 ?

michi-covalent:
@Rolinh yes, thanks. just closed now.

Issue #155, opened 2020-02-27
Numerical identity for endpoints not always populated


gandro:
Some Cilium trace notifications (e.g. `from-endpoint` flows) do not contain the numerical security identity of the source/target (in the case of `from-endpoint`, this is more or less by design, since it is intended to be an early tracepoint that happens before any lookups/modifications).

For source/targets that are not a local endpoint, this is not a problem as the identities are populated when we perform the IPCache lookup. For local endpoints however, we fail to perform a lookup to obtain this information, since our `EndpointGetter` does not have field for the security identity. 

We should extend our new `EndpointInfo` interface to expose the numerical security identity of the endpoint, so we can populate the field in the parser.

PR #157, opened 2020-02-27
 parser: Use numeric security identity from endpoint

gandro:
This adds the numerical security identity to `v1.Endpoint` and the `EndpointInfo` interface. The `GetIdentity` function is also implemented by Cilium's `endpoint.Endpoint` struct (required for embedding).

The parser is modified use this newly available numeric security identity from the endpoint getter instead of relying on information from the `TraceNotify` struct.

This fixes a bug where the security identity was missing, because `TraceNotify` does not always contain the actual security identity (for example for `TRACE_FROM_LXC` events).

Fixes: #155

### DIFF
--- a/pkg/parser/threefour/parser.go
+++ b/pkg/parser/threefour/parser.go
@@ -231,7 +231,7 @@ func (p *Parser) resolveEndpoint(ip net.IP, securityIdentity uint64) *pb.Endpoin
 		if ep, ok := p.endpointGetter.GetEndpointInfo(ip); ok {
 			return &pb.Endpoint{
 				ID:        ep.GetID(),
-				Identity:  securityIdentity,
+				Identity:  uint64(ep.GetIdentity()),
 				Namespace: ep.GetK8sNamespace(),
 				Labels:    sortAndFilterLabels(ep.GetLabels(), securityIdentity),
 				PodName:   ep.GetK8sPodName(),
--- a/pkg/api/v1/interface.go
+++ b/pkg/api/v1/interface.go
@@ -15,6 +15,7 @@
 package v1
 
 import (
+	"github.com/cilium/cilium/pkg/identity"
 	pb "github.com/cilium/hubble/api/v1/flow"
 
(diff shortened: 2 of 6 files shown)


pkg/parser/threefour/parser.go
```diff
@@ -231,7 +231,7 @@ func (p *Parser) resolveEndpoint(ip net.IP, securityIdentity uint64) *pb.Endpoin
 		if ep, ok := p.endpointGetter.GetEndpointInfo(ip); ok {
 			return &pb.Endpoint{
 				ID:        ep.GetID(),
-				Identity:  securityIdentity,
+				Identity:  uint64(ep.GetIdentity()),
```
Rolinh: Why is `Endpoint.Identity` defined as a `uint64` and not a `uint32` like in `identity.NumericIdentity` ? At first glance, it seems that these type casts could be avoided.

gandro: Good question. Unfortunately, neither Cilium nor Hubble are very consistent here (the `Endpoint.ID` is also a `uint64` in the interfaces and protobuf definitions, but internally and in the datapath a 16bit unsigned).

What we can do is change the data type of pb.Endpoint to `uint32`. It would be a breaking protobuf change though, something we should do before a proper release.

gandro: Ah, thinking about it, the reason is probably that the auto-generated OpenAPI definitions from Cilium use `int64` for any `integer` types, so we adapted this without checking the actual data size. We should revisit the integer types we use in the protobuf and make sure they are appropriately sized.

gandro: Created https://github.com/cilium/hubble/issues/158

michi-covalent:
thanks @gandro let me review this after the meeting today

gandro:
> thanks @gandro let me review this after the meeting today

Sure! Note that this PR contains a fix for #155 for both stand-alone Hubble and embedded Hubble.

PR #159, opened 2020-02-27
container: add a priority queue implementation for events

Rolinh:
This priority queue will be used for multi-node support in order to be able to re-order events in chronological order as they arrive.

Updates #89

### DIFF
--- a/pkg/container/priority_queue.go
+++ b/pkg/container/priority_queue.go
@@ -1,0 +1,95 @@
+// Copyright 2020 Authors of Hubble
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package container
+
+import (
+	"container/heap"
+
+	"github.com/cilium/hubble/pkg/api/v1"
+)
+
+// PriorityQueue is a priority queue of events that implements heap.Interface.
+// Events are ordered based on their timestamp value, the oldest event having
+// the higher priority.
+//
+// This implementation is loosely based on the priority queue example in
+// "container/heap".
+type PriorityQueue struct {
+	h minHeap
+}
+
+type minHeap []*v1.Event
+
+// Ensure that minHeap implements heap.Interface.
+var _ heap.Interface = (*minHeap)(nil)
+
+// NewPriorityQueue creates a new PriorityQueue with initial capacity n.
+func NewPriorityQueue(n int) *PriorityQueue {
+	h := make(minHeap, 0, n)
+	heap.Init(&h)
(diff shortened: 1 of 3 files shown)


Rolinh:
@gandro Just fixed doc comments with these last 2 pushes.

Rolinh:
> i guess there is no existing priority queue implementation we can just use?

Do you want to bring in the topic of Go's lack of generics to the table? :)

michi-covalent:
ok ok i'll wait for go 2 or something :)

pkg/container/priority_queue.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+// PriorityQueue is a priority queue of events that implements heap.Interface.
+// Events are ordered based on their timestamp value, the oldest event having
+// the higher priority.
+//
+// This implementation is loosely based on the priority queue example in
+// "container/heap".
+type PriorityQueue struct {
+	pq priorityQueue
```
glibsm: priority queue embeds a priority queue? some choice of names here is odd

pkg/container/priority_queue.go
```diff
@@ -0,0 +1,103 @@
...
+}
+type priorityQueue []*v1.Event
+
+// Ensure that priorityQueue implements heap.Interface.
+var _ heap.Interface = (*priorityQueue)(nil)
+
+// NewPriorityQueue creates a new PriorityQueue with initial capacity n.
+// It may optionally be initialized with events in which case the initial
+// capacity is set to n or the number of events, whichever is greater.
+func NewPriorityQueue(n int, events ...*v1.Event) *PriorityQueue {
```
glibsm: i woudln't commit varargs to the list of events, i'd either use the options pattern (and a list of events as one of the options), or just omit it all together

pkg/container/priority_queue.go
```diff
@@ -0,0 +1,103 @@
...
+	for i, e := range events {
+		pq[i] = e
+	}
+	heap.Init(&pq)
+	return &PriorityQueue{pq}
+}
+
+// Len is the number of events in the queue.
+func (pq PriorityQueue) Len() int {
+	return pq.pq.Len()
```
glibsm: `pq.pq` makes me thing you should either embed the inner one, or name it `list` or something i dunno

Rolinh: By embedding the inner one, it would expose methods required to implement `heap.Interface` such as `Swap()`. It would also force `Pop()`/`Push()` to use `interface{}` and thus impose the caller to type cast each time. I think neither of this is desirable.
If I name the inner one `list`, it would be wrong because it's not simply a least. It could be named something like `heap` if you think it would be better.

pkg/container/priority_queue.go
Rolinh: Using the option pattern seems overkill for this use-case. However, I have nothing against omitting it altogether as, all things considered, I don't think it's really required/useful anyway.

Rolinh:
@glibsm PTAL

glibsm:
Thoughts on moving this to `internal/*`? pkg is kind of an old pattern, as discussed recently, and this is very specific for hubble that we don't want other things importing

Rolinh:
> Thoughts on moving this to `internal/*`? pkg is kind of an old pattern, as discussed recently, and this is very specific for hubble that we don't want other things importing

I'm all for it! I think that actually much more code than just this could be moved to `internal` but I'd do this as part of another PR.

Issue #161, opened 2020-02-27
`GetEventsChannel` buffer size too small embedded Hubble


gandro:
The current default is too small for embedded Hubble. From https://github.com/cilium/cilium/pull/10238/files#r385174692:

> The issue seems to be that `GetEventsChannel()` buffer size is too small for the "Embedded Mode": By default, the channel returned `GetEventsChannel` will only store 81 flows (1% of the next power of two of `--max-flows`). This queue size is way to small to deal with short bursts.
> 
> The previous buffer size (in Hubble stand-alone mode) has been `GetEventsChannel()` (up to 100 flows) + Cilium monitor's socket queue, which is `1024*NumberOfCPUs` (capped at `16384`).

gandro:
Note: While the Hubble side has been merged, the Cilium change still has to be done.

michi-covalent:
i'll take care of it

gandro:
Closed via https://github.com/cilium/cilium/pull/10488

PR #163, opened 2020-02-27
server: Make event queue size configurable

gandro:
When Hubble runs in embedded mode inside Cilium, the event queue size
must be big enough so the Hubble flow decoder can keep up with short
bursts of flows. The current calculation of 1% of the ring buffer size
has proven to be too small for embedded mode. For comparision, Cilium
monitor uses a event queue size of 1024 * NumberOfCPUs.

However, when Hubble runs as its own process reading from the Cilium
monitor socket, the Cilium monitor queue already acts as this buffer.
Therefore the Hubble event size is less critical. In fact, it was
capped at 100. This PR simplifies the logic and sets the default for
stand-alone mode to 128.

The value for embedded mode will be configurable in cilium-agent.

Ref: #161

### DIFF
--- a/pkg/server/local_observer.go
+++ b/pkg/server/local_observer.go
@@ -20,7 +20,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/cilium/cilium/pkg/math"
 	pb "github.com/cilium/hubble/api/v1/flow"
 	"github.com/cilium/hubble/api/v1/observer"
 	v1 "github.com/cilium/hubble/pkg/api/v1"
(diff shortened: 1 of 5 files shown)


gandro:
cc @michi-covalent This is related, but imo not a blocker, to embedded Hubble

pkg/server/observer.go
```diff
@@ -60,12 +60,13 @@ func NewServer(
 	serviceCache *servicecache.ServiceCache,
 	payloadParser *parser.Parser,
 	maxFlows int,
+	eventQueueSize int,
```
Rolinh: I'm not really comfortable with adding yet another parameter to `NewServer()` as the list is already quite long. Eventually, I think this should be refactored.

pkg/server/local_observer.go
```diff
@@ -81,13 +80,14 @@ type LocalObserverServer struct {
 func NewLocalServer(
 	payloadParser *parser.Parser,
 	maxFlows int,
+	eventQueueSize int,
 	logger *logrus.Entry,
 ) *LocalObserverServer {
 	return &LocalObserverServer{
 		log:  logger,
 		ring: container.NewRing(maxFlows),
 		// have a channel with 1% of the max flows that we can receive
```
glibsm: comment is now wrong

pkg/server/local_observer.go
```diff
@@ -81,13 +80,14 @@ type LocalObserverServer struct {
...
-		events:        make(chan *pb.Payload, uint64(math.IntMin(maxFlows/100, 100))),
+		events:        make(chan *pb.Payload, eventQueueSize),
```
glibsm: is `make(chan, 0)` === `make(chan)`?

michi-covalent: yes, it returns an unbuffered channel.

pkg/server/local_observer.go
gandro: Removed!

pkg/server/observer.go
glibsm: yes. Options coming soon

PR #162, opened 2020-02-27
Fix trivial typo on README

pedro823:
I was reading the readme, found this simple typo.

I don't think any explanation's necessary :)

PR #165, opened 2020-02-28
add version information

Rolinh:
This PR adds a `--version` flag to the root command:
```
$ hubble --version
hubble v0.0.0
```
As well as a `version` subcommand for more detailed version information:
```
$ hubble version
hubble v0.0.0 compiled with go1.14 on linux/amd64
```

This command is meant to provide more information in the future such as the commit hash, the server version/mode, etc.

### DIFF
--- a/pkg/version.go
+++ b/pkg/version.go
@@ -1,0 +1,18 @@
+// Copyright 2020 Authors of Hubble
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


pkg/version.go
```diff
@@ -0,0 +1,22 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pkg
+
+const (
+	// Name is the software name.
+	Name = "Hubble"
```
glibsm: name doesn't seem that necessary

Rolinh: I guess I fear that ["Hublbe"](https://github.com/cilium/hubble/pull/162#issuecomment-592102864) could make a comeback. :)
I'm totally fine removing it though.

PR #167, opened 2020-03-02
k8s: add high-level kube apiserver client

Rolinh:
The goal is to be able to retrieve the list of hubble peers using the kubernetes API. This will be useful to implement the multi-node feature. 

ref #89


michi-covalent:
so i guess we need to talk to k8s api after all?

Rolinh:
@michi-covalent Yup, for non-embedded mode. I'll define an interface though so one can implement it using the k8s client for non-embedded mode and using cilium node manager when embedded.
There are other PRs in the work ;)

michi-covalent:
guess this is not related
```
--- FAIL: TestRingReader_NextFollow (0.00s)
    --- FAIL: TestRingReader_NextFollow/read_2,_start_at_position_0,_expect_timeout=false (0.00s)
        ring_reader_test.go:219: 
            	Error Trace:	ring_reader_test.go:219
            	Error:      	Not equal: 
            	            	expected: []*v1.Event{(*v1.Event)(0xc000521180), (*v1.Event)(0xc0005211a0)}
            	            	actual  : []*v1.Event{(*v1.Event)(0xc000520f20), (*v1.Event)(nil)}
            	            	
            	            	Diff:
            	            	--- Expected
```

Rolinh:
> lgtm, although i feel like we should go all in with the embedded mode and deprecate the standalone mode altogether.

What about support for Cilium < 1.8?

Rolinh:
Hubble embedded will be a requirement for the upcoming multi-node feature in order to reduce code complexity and to avoid overloading the kube apiserver. Thus, let's close this PR.

Rolinh:
Re-opening. All things considered, this will be useful when running a hubble client that is not necessarily embedded into cilium in order to discover hubble peers.

Rolinh:
Will close for now, this might be implemented differently.

PR #170, opened 2020-03-03
vendor: Remove gojay

michi-covalent:
This PR removes the WithJSONEncoder printer option and gojay dependency
since the observe command is always configured to use the standard JSON
encoder.


glibsm:
The fact that no tests broke means one of two things:
1. There weren't tests 😄 
2. They still pass

I'm assuming the latter?

michi-covalent:
> The fact that no tests broke means one of two things:
> 
>     1. There weren't tests 😄
> 
>     2. They still pass
> 
> 
> I'm assuming the latter?

definitely the latter. the observe command never used gojay encoder :)

vendor/modules.txt
```diff
@@ -63,7 +62,6 @@ github.com/davecgh/go-spew/spew
 github.com/fatih/color
 # github.com/francoispqt/gojay v1.2.13
 ## explicit
```
Rolinh: Mmh, this shouldn't be an explicit dependency anymore. Please, re-run `go mod tidy && go mod vendor`.

michi-covalent: done. somehow i always keep forgetting this :(

michi-covalent:
will remove the interface

PR #174, opened 2020-03-05
ring: fix race of ring between goroutines

Cordius:
GRPC server for client-listener read flows from the ring on the
one routine, while the GRPC server handling the events from Cilium
Monitor write to the the same ring on the other goroutine. That may
incur race, this patch fix it.

Signed-off-by: Wu Zongyong <wuzongyong@linux.alibaba.com>

### DIFF
--- a/pkg/container/ring.go
+++ b/pkg/container/ring.go
@@ -45,6 +45,9 @@ type Ring struct {
 	// ReadFrom(<-chan struct{}, uint64) <-chan *pb.Payload knows the writer
 	// has written into the internal buffer.
 	cond *sync.Cond
+	// mutex is used to prevent races when different goroutines read from or
+	// write into the internal buffer.
+	mutex sync.Mutex
 }
 
 // NewRing creates a ring buffer. For efficiency, the internal
@@ -71,6 +74,7 @@ func NewRing(n int) *Ring {
 		dataLen:   dataLen,
 		data:      make([]*v1.Event, dataLen, dataLen),
 		cond:      sync.NewCond(&sync.RWMutex{}),
+		mutex:     sync.Mutex{},
 	}
 }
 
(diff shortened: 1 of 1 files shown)


Rolinh:
We're aware of this issue and track it in #82. However, the ring buffer is a hot path and we would thus ideally like to keep it lock free. The idea is that if we ensure that the reads happen at a different index that the writes, then a lock is not required. Obviously, there is a race in the current implementation but it should be possible to fix it without introducing a lock.

By introducing a lock, the ring gets slower. This can be verified by running the benchmark:
```
go test -timeout=30s -bench='BenchmarkRing*' github.com/cilium/hubble/pkg/container
```

Here is the result before this patch on my workstation:
```
BenchmarkRingWrite-32          	177003388	         6.54 ns/op	       0 B/op	       0 allocs/op
BenchmarkRingRead-32           	760685061	         2.04 ns/op	       0 B/op	       0 allocs/op
```

And after the introduction of the lock in this PR:
```
BenchmarkRingWrite-32          	86751697	        14.0 ns/op	       0 B/op	       0 allocs/op
BenchmarkRingRead-32           	121590757	        10.3 ns/op	       0 B/op	       0 allocs/op
```
So based on this benchmark ring read seem to be ~5x slower and write ~2x.

pkg/container/ring.go
```diff
@@ -71,6 +74,7 @@ func NewRing(n int) *Ring {
 		dataLen:   dataLen,
 		data:      make([]*v1.Event, dataLen, dataLen),
 		cond:      sync.NewCond(&sync.RWMutex{}),
+		mutex:     sync.Mutex{},
```
Rolinh: The zero value for a Mutex is an unlocked mutex, thus this is not required.

Rolinh:
Based on the performance metrics, let's close this PR in favor of #177.

PR #176, opened 2020-03-05
Fix go vet reports

Rolinh:
Details in individual commit messages.


Rolinh:
Nope, not anymore. We should probably also configure `go vet` to run as part of the standard checks. Lemme create a PR to add this.

PR #177, opened 2020-03-05
container/ring: Fix various data races and race conditions

gandro:
This is a first step towards fixing #82. This uses
`atomic.{Load,Store}Pointer` to access the individual slots in the ring
buffer, ensuring that each write is properly propagated to all other
threads. While it seems to satisfy the Go race detector, it seems that
running the unit test now with `-race` will now sometimes fail with a
semantic error, i.e. the ring buffer content is not what is expected.
This indicates that there remains a real race condition somewhere, this
commit only addresses the data race discovered by the Go race detector.

This means that #142 remains blocked, as we also have to find the root
cause for the race condition. Tests pass consistently when running
this commit without `-race` however, therefore it does not seem to
introduce any regressions.

### DIFF
--- a/pkg/container/ring.go
+++ b/pkg/container/ring.go
@@ -108,7 +108,8 @@ func (r *Ring) Cap() uint64 {
 }
 
 // Write writes the given event into the ring buffer in the next available
-// writing block.
+// writing block. The entry must not be nil, otherwise readFrom will block when
+// reading back this event.
 func (r *Ring) Write(entry *v1.Event) {
 	write := atomic.AddUint64(&r.write, 1)
 	writeIdx := (write - 1) & r.mask
@@ -202,7 +203,7 @@ func (r *Ring) readFrom(ctx context.Context, read uint64) <-chan *v1.Event {
 			// write: f0 f1 f2 f3 f4 f5 f6 f7 f8 f9 fa fb fc fd fe ff  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
 			// index:  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
 			// cycle: 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
-			case readCycle == (writeCycle-1)&r.cycleMask && readIdx > lastWriteIdx:
+			case event != nil && readCycle == (writeCycle-1)&r.cycleMask && readIdx > lastWriteIdx:
 				select {
 				case ch <- event:
 					continue
@@ -216,7 +217,7 @@ func (r *Ring) readFrom(ctx context.Context, read uint64) <-chan *v1.Event {
 			// write: f0 f1 f2 f3 f4 f5 f6 f7 f8 f9 fa fb fc fd fe ff  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
 			// index:  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
 			// cycle: 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
-			case readCycle == writeCycle:
+			case event != nil && readCycle == writeCycle:
 				if readIdx < lastWriteIdx {
 					select {
 					case ch <- event:
@@ -243,7 +244,7 @@ func (r *Ring) readFrom(ctx context.Context, read uint64) <-chan *v1.Event {
 			// write: f0 f1 f2 f3 f4 f5 f6 f7 f8 f9 fa fb fc fd fe ff  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
 			// index:  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
 			// cycle: 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f 1f  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
-			case readCycle >= (writeCycle+1)&r.cycleMask && readCycle < (halfCycle+writeCycle)&r.cycleMask:
+			case event == nil || readCycle >= (writeCycle+1)&r.cycleMask && readCycle < (halfCycle+writeCycle)&r.cycleMask:
 				// The writer has already written a new event so there's no
 				// need to stop the reader.
 				//
--- a/pkg/container/ring.go
+++ b/pkg/container/ring.go
@@ -42,10 +42,11 @@ type Ring struct {
 	dataLen uint64
 	// data is the internal buffer of this ring buffer.
 	data []*v1.Event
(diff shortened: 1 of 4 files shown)


gandro:
Performance impact on my laptop:

Before:
```
goos: linux
goarch: amd64
pkg: github.com/cilium/hubble/pkg/container
BenchmarkRingWrite-8   	116604054	        10.1 ns/op	       0 B/op	       0 allocs/op
BenchmarkRingRead-8    	160090080	        12.8 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/cilium/hubble/pkg/container	10.292s

```

After:
```
goos: linux
goarch: amd64
pkg: github.com/cilium/hubble/pkg/container
BenchmarkRingWrite-8   	72918658	        16.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkRingRead-8    	287611406	        18.1 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/cilium/hubble/pkg/container	22.051s
```

gandro:
Seems like the failed unit test that I discovered is a flake/timeout which only occurs when `-race` is enabled. I guess that's expected, since the Go race detector can impose quite the overhead. See https://github.com/cilium/hubble/pull/177/commits/35ff424ec9bd499977c0ca9a9c36a78d77f0ec22

This means that with this PR we likely can merge #142  and close #82.

pkg/container/ring.go
```diff
@@ -16,11 +16,11 @@ package container
 
 import (
 	"context"
-	"sync"
-	"sync/atomic"
-
 	v1 "github.com/cilium/hubble/pkg/api/v1"
 	"github.com/cilium/hubble/pkg/math"
+	"sync"
```
Rolinh: What's the reason for moving stdlib imports there?

Rolinh:
I've just hit a ring reader test failure with these changes:
```
--- FAIL: TestRingReader_NextFollow (0.32s)
    --- FAIL: TestRingReader_NextFollow/read_5,_start_at_position_5,_expect_timeout=false (0.22s)
        ring_reader_test.go:219: 
            	Error Trace:	ring_reader_test.go:219
            	Error:      	Not equal: 
            	            	expected: []*v1.Event{(*v1.Event)(0xc000537880), (*v1.Event)(0xc0005378a0), (*v1.Event)(0xc0005378c0), (*v1.Event)(0xc0005378e0), (*v1.Event)(0xc000537920)}
            	            	actual  : []*v1.Event{(*v1.Event)(0xc000537640), (*v1.Event)(0xc000537660), (*v1.Event)(0xc0005376a0), (*v1.Event)(nil), (*v1.Event)(nil)}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -13,10 +13,4 @@
            	            	  }),
            	            	- (*v1.Event)({
            	            	-  Timestamp: (*types.Timestamp)(1970-01-01T00:00:08Z),
            	            	-  Event: (interface {}) <nil>
            	            	- }),
            	            	- (*v1.Event)({
            	            	-  Timestamp: (*types.Timestamp)(1970-01-01T00:00:09Z),
            	            	-  Event: (interface {}) <nil>
            	            	- })
            	            	+ (*v1.Event)(<nil>),
            	            	+ (*v1.Event)(<nil>)
            	            	 }
            	Test:       	TestRingReader_NextFollow/read_5,_start_at_position_5,_expect_timeout=false
        ring_reader_test.go:220: 
            	Error Trace:	ring_reader_test.go:220
            	Error:      	Not equal: 
            	            	expected: false
            	            	actual  : true
            	Test:       	TestRingReader_NextFollow/read_5,_start_at_position_5,_expect_timeout=false
FAIL
```

gandro:
I have identified a few race conditions and believe to have fixed them. See commit messages for detail. I have not observed any data race anymore, so I'm confident that this fixes #82.

pkg/container/ring_reader.go
```diff
@@ -64,20 +66,28 @@ func (r *RingReader) Next() *v1.Event {
...
-	if r.c == nil {
+	// if the context changed between invocations, we also have to restart
+	// readFrom, as the old readFrom instance will be using the old context.
+	if r.c == nil || r.ctx != ctx {
 		r.c = r.ring.readFrom(ctx, r.idx)
+		r.ctx = ctx
 	}
-	// when the ring is not full, nil events may be sent to the channel
-	// one should keep reading in such a case
 	for {
```
Rolinh: If the `ring.readFrom()` does not return `nil` events anymore (it blocks until it can read a non-nil event), then this for loop his not needed anymore.

pkg/container/ring.go
```diff
@@ -224,37 +260,38 @@ func (r *Ring) readFrom(ctx context.Context, read uint64) <-chan *v1.Event {
...
-					r.cond.L.Unlock()
-					return
-				default:
-					r.cond.Wait()
-					r.cond.L.Unlock()
-					read--
+
+				// This channel will be closed by the writer if it makes a write
+				if r.notifyCh == nil {
+					r.notifyCh = make(chan struct{})
```
Rolinh: Isn't possible to simply send an empty struct into `notifyCh` rather than re-creating the channel every time?

gandro: Unfortunately not. We need to close the channel since close is broadcast to all consumers. A simple write would only notify a single consumer.

Rolinh: Then wouldn't [`sync.Cond`](https://golang.org/pkg/sync/#Cond) be more appropriate as, in effect, you want to broadcast to all consumers?

gandro: This is what we had previously. Unfortunately it is not flexible enough to react to context cancellation (this seems to be a common issue, see https://github.com/golang/go/issues/16620):

From the commit message:
```
The usage of `sync.Cond` in `Ring.readFrom` was prone to races, i.e.
it could miss signals that should stop it from going to sleep.
The following two cases addressed by this commit:

1. The `readFrom` goroutine could to sleep right after `r.write` has
   been bumped, therefore missing the signal from `cond.Broadcast()`.
   The previous code has tried to mitigate this (see removed FIXME
   comment), however that check still suffered from TOCTOU and
   therefore was not sufficient. This is now addressed by using a mutex
   in the writer that is held between the time where `r.write` is
   updated and the signal is being sent. This mutex is not held by any
   readers which are not trying to read the current write pointer,
   therefore concurrency of readers with the writer is preserved for
   most cases.

2. The `readFrom` goroutine could also enter sleep without noticing
   that its context is cancelled right before `cond.Wait()` is called.
   While there was a goroutine which sent a signal when the context is
   cancelled, there was still a possibility for a race where this
   signal was sent before `cond.Wait()` was actually called. This is
   mitigated by replacing `sync.Cond` with a channel based notification
   mechanism that allows us to use a `select` statement to wait for both
   the writer notification and the context cancellation.
```

pkg/container/ring_reader.go
gandro: Good catch! Fixed!

pkg/container/ring.go
```diff
@@ -89,12 +110,30 @@ func (r *Ring) Cap() uint64 {
...
+
 	write := atomic.AddUint64(&r.write, 1)
 	writeIdx := (write - 1) & r.mask
-	r.data[writeIdx] = entry
-	r.cond.Broadcast()
+	r.dataStoreAtomic(writeIdx, entry)
+
+	// notify any sleeping readers
+	if r.notifyCh != nil {
+		close(r.notifyCh)
```
Rolinh: Is it allowed to have multiple writers? Because if yes, then we have a race here and the channel could be closed twice thus leading to panic.

pkg/container/ring.go
Rolinh: OK let's do it this way.

pkg/container/ring.go
gandro: Yeah... also we don't allow multiple writers so far :)

PR #178, opened 2020-03-05
make: add vet target

Rolinh:


.github/workflows/unittests.yml
```diff
@@ -23,4 +23,4 @@ jobs:
         export PATH=${PATH}:${GOPATH}/bin
         GO111MODULE=off go get -u github.com/gordonklaus/ineffassign
         GO111MODULE=off go get -u golang.org/x/lint/golint
-        make check-fmt lint test
+        make lint vet test
```
michi-covalent: no more ./contrib/scripts/check-fmt.sh?

Rolinh: It's a `lint` dep so it was effectively run twice. I think just once is enough :)

michi-covalent: maybe it'll catch something the second time :)

.github/workflows/unittests.yml
```diff
@@ -23,4 +23,4 @@ jobs:
         export PATH=${PATH}:${GOPATH}/bin
         GO111MODULE=off go get -u github.com/gordonklaus/ineffassign
         GO111MODULE=off go get -u golang.org/x/lint/golint
-        make check-fmt lint test
```
glibsm: why was check-fmt removed?

Rolinh: See ab53f3f2b507eff8210e0edcf5bc784038b28563 or https://github.com/cilium/hubble/pull/178#discussion_r388561406

glibsm: Then why wasn't that file deleted? I wouldn't have any questions then :)

Rolinh: No file was deleted as part of this PR :thinking:

michi-covalent: seems like this got more confusing than necessary :) maybe we should have a single target that does all the formatting/linting/vet checks.

Rolinh: @michi-covalent See https://github.com/cilium/hubble/pull/181

Issue #76, opened 2020-02-04
Remove endpoint history from endpoint cache


gandro:
The current endpoint cache in `github.com/cilium/hubble/pkg/api/v1.Endpoints` contains a few artifacts from early versions of Hubble where flows were decoded late, i.e. when a gRPC flow requested them. This meant that the endpoint cache had to keep "historic" data about endpoints around, we do this by tracking its deletion time:

https://github.com/cilium/hubble/blob/9df73911732a4d880c4fb4c7d112e45b4ed28fa0/pkg/api/v1/types.go#L29

Nowadays however, endpoint information is annotated as soon as a flow is observed, making it pointless to keep deleted endpoints around. 

We should refactor the endpoint cache and get rid of the "deleted endpoints" functionality, as it complicates the code (every getter has to check that endpoints are not deleted). This also gives us the opportunity to fix a few other issues with the current package, such as:

- Its location in `api/v1`, even though it is not really related to the gRPC interface anymore. It should probably look more similar to the fqdn/ip/service caches.
- The potential footguns it has with regards to concurrency, i.e. it returns pointers whose access is supposed to see synchronized (see https://github.com/cilium/hubble/pull/74#discussion_r374572739)

Rolinh:
This issue is no longer relevant now that hubble server is embedded into the cilium agent.

PR #179, opened 2020-03-06
api/endpoint: stop marking endpoints as deleted, just remove them

Rolinh:
The endpoint cache contained a few artifacts from early versions of
Hubble where flows were decoded late, i.e. when a gRPC flow requested
them. This meant that the endpoint cache had to keep "historic" data
about endpoints around. This was done by tracking endpoints deletion
time using Endpoint's Deleted field.

As now now, however, endpoint information is annotated as soon as a flow
is observed, making it pointless to keep deleted endpoints around.

This commit removes the Deleted field from the Endpoint struct and
adjust the code so that when an endpoint has to be deleted, it is
actually deleted rather than soft-deleted then garbage collected.
This also removes the need to garbage collect soft-deleted endpoints
from the endpoint cache.

Ref: #76

### DIFF
--- a/pkg/api/v1/endpoint.go
+++ b/pkg/api/v1/endpoint.go
@@ -16,17 +16,15 @@ package v1
 
 import (
 	"net"
-	"time"
 )
 
 // EndpointsHandler defines an interface for interacting with Cilium endpoints.
 type EndpointsHandler interface {
 	SyncEndpoints([]*Endpoint)
 	UpdateEndpoint(*Endpoint)
-	MarkDeleted(*Endpoint)
 	FindEPs(epID uint64, ns, pod string) []Endpoint
 	GetEndpoint(ip net.IP) (endpoint *Endpoint, ok bool)
-	GarbageCollect()
+	DeleteEndpoint(*Endpoint)
 	GetEndpointByContainerID(id string) (*Endpoint, bool)
 	GetEndpointByPodName(namespace string, name string) (*Endpoint, bool)
 }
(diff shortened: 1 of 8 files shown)


glibsm:
Given @gandro did the initial work and created the cleanup ticket, I'd like him to sign off on this before this is merged

pkg/api/v1/endpoint.go
```diff
@@ -92,44 +86,33 @@ func (e *Endpoint) DeepCopy() *Endpoint {
...
-			ep.Deleted = &t
-		}
-	}
-
 	// Add the endpoint to the list of endpoints.
 	for _, updatedEp := range newEps {
 		es.updateEndpoint(updatedEp)
 	}
+	// some endpoints were deleted, remove them
+	if len(es.eps) != len(newEps) {
```
gandro: I don't think this if statement is correct, the length itself is not sufficient to detect deletion. If `es.eps=={1,2,3}` and `newEps=={1,4,5}`, this branch would never be taken even though we have two deleted endpoints (`2` and `3`) that must be removed via `es.deleteEndpoint`.

Rolinh: The lines above actually add new endpoints. So once you reach line `102`, you'd actually have `es.eps == {1,2,3,4,5}` which is of a different size than `newEps == {1,4,5}`. So in fact, the branch would be taken.

gandro: Ah! Yes, you are right, I missed that. Sorry for the noise :)

Rolinh:
> probably not directly related, but i don't think we use the `Created` field either.

See https://github.com/cilium/hubble/pull/186

PR #181, opened 2020-03-09
add make check target; separate step for tests

Rolinh:
- make: add check target
- actions: use `make check` for static checks, add separate step for tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,11 +26,7 @@ test:
 bench:
 	go test -timeout=30s -bench=. $$(go list ./...)
 
-lint: check-fmt ineffassign
-ifeq (, $(shell which golint))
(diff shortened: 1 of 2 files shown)


michi-covalent:
let's roll

Makefile
```diff
@@ -41,10 +37,16 @@ ifeq (, $(shell which ineffassign))
 endif
 	ineffassign .
 
+lint:
+ifeq (, $(shell which golint))
+	$(error "golint not installed; you can install it with `go get -u golang.org/x/lint/golint`")
```
glibsm: note for future: we can vendor golint through go mod and run it from inside the vendor dir so people don't have to have a separate binary (or incompatible version)

PR #182, opened 2020-03-09
cmd: fix teardown

Rolinh:
See individual commits for details.


cmd/root.go
```diff
@@ -31,28 +33,59 @@ import (
...
 // This is called by main.main(). It only needs to happen once to the rootCmd.
-func Execute() {
-	fin := maybeProfile()
-	defer fin() // make sure update memory profile is written in the end
-
-	if err := rootCmd.Execute(); err != nil {
+func Execute() error {
+	err := rootCmd.Execute()
+	if err != nil {
 		fmt.Println(err)
```
glibsm: it's better to return the error and let the caller choose print or not

main.go
```diff
@@ -19,9 +19,13 @@
...
 import (
+	"os"
+
 	"github.com/cilium/hubble/cmd"
 )
 
 func main() {
-	cmd.Execute()
+	if err := cmd.Execute(); err != nil {
+		os.Exit(-1)
```
glibsm: log.Fatal is probably more appropriate with the error message

cmd/root.go
```diff
@@ -31,28 +33,59 @@ import (
...
-	Use:     "hubble",
-	Short:   "CLI",
-	Long:    `Hubble is a utility to observe and inspect recent Cilium routed traffic in a cluster.`,
-	Version: pkg.Version,
+	Use:          "hubble",
+	Short:        "CLI",
+	Long:         `Hubble is a utility to observe and inspect recent Cilium routed traffic in a cluster.`,
+	SilenceUsage: true,
+	Version:      pkg.Version,
+	PersistentPreRunE: func(_ *cobra.Command, _ []string) error {
```
glibsm: maybe move the inners to a separate function - the command definition will the remain cleaner

cmd/serve/serve.go
```diff
@@ -101,12 +102,14 @@ func New(log *logrus.Entry) *cobra.Command {
 				int(eventQueueSize),
 				log,
 			)
+			ctx, cancel := context.WithCancel(context.Background())
+			setupSigHandler(ctx, cancel)
```
glibsm: i don't think cancel is called in all cases with this setup (crashing or other error without ctrl-c)

.gitignore
```diff
@@ -1,2 +1,3 @@
 /hubble
 install/kubernetes/hubble.yaml
+*.pprof
```
glibsm: we don't need to remove it from here, but this is a fantastic addition to your global gitignore so pprof files are never checked in across all projects

cmd/serve/serve.go
```diff
@@ -243,23 +243,24 @@ func setupListeners(listenClientUrls []string) (listeners map[string]net.Listene
...
-				"Shutting down after the configured duration",
-			)
-			os.Exit(0)
-		}()
+		go func(ctx context.Context) {
+			select {
+			case <-time.After(serveDuration):
+				log.WithField("duration", serveDuration).Info(
+					"Shutting down after the configured duration",
+				)
```
Rolinh: Just realized that the context should be cancelled here. Do we want to keep the server shutdown after duration feature though? What's the use-case for this?

gandro: I believe the `serveDuration` was introduced to make integration tests simpler, but the integration tests never landed. However, given that we now have embedded Hubble in Cilium, this the `serveDuration` is becoming less and less useful, as even if we stop the Hubble server, `cilium-agent` will continue to run.

I agree with @michi-covalent (https://github.com/cilium/hubble/pull/182#pullrequestreview-371363840) that we should simply remove `serveDuration`.

Rolinh: OK, I'll take care of this.

Rolinh:
> i kind of want to just get rid of serveDuration. i don't think anybody's using it.

See https://github.com/cilium/hubble/pull/185

PR #183, opened 2020-03-09
pkg/server: Option

glibsm:
Two individual commits. One adds the pattern for options, the other implements the numerical options for the ring size and monitor buffer.


pkg/server/local_observer_test.go
```diff
@@ -47,14 +50,18 @@ func TestNewLocalServer(t *testing.T) {
 }
 
 func TestLocalObserverServer_ServerStatus(t *testing.T) {
+	// (glibsm): This test is really confusing. `serveroption.WithMaxFlows(1)`
```
michi-covalent: yeah i guess this comes from the fact that the ring buffer size gets rounded up to the next power of 2.

it would have been less confusing if the parameter actually specified the power instead of implicitly rounding the number up inside the constructor, but what can we do.

PR #188, opened 2020-03-12
metrics: Export EnableMetrics

michi-covalent:
2 commits:

- pick up the latest cilium
- export EnableMetrics function. i forgot it was being used by embedded mode.


Rolinh:
Ah, my bad, I had no idea it was being used by hubble embedded as I didn't expect anything from `cmd` to be imported elsewhere.

PR #189, opened 2020-03-12
helm: Add priorityClassName for kube-system only

michi-covalent:
Don't mark Hubble pod as system-node-critical if it's not in kube-system
namespace. Later versions of Kubernetes don't have this restriction, but
for now I'm simply setting this field based on which namespace Hubble is
in.

Ref: https://github.com/kubernetes/kubernetes/pull/76310
Ref: https://github.com/cilium/cilium/issues/10504


michi-covalent:
i don't even know if hubble should ever be marked as system-node-critical.

PR #190, opened 2020-03-17
cmd: remove syscall references

Rolinh:
The `syscall` package is deprecated since Go 1.4:

> Deprecated: this package is locked down. Callers should use the
> corresponding package in the golang.org/x/sys repository instead. That
> is also where updates required by new systems or versions should be
> applied. See https://golang.org/s/go1.4-syscall for more information.

In effect the `syscall` package is not updated, "not even to keep pace with changes in operating systems it references".

This commit replaces references to the `syscall` package by `golang.org/x/sys/unix`, as hinted by the Go documentation of the `syscall` package. This has already been done on Cilium (see https://github.com/cilium/cilium/issues/10116).

### DIFF
--- a/pkg/printer/options.go
+++ b/pkg/printer/options.go
@@ -16,8 +16,7 @@ package printer
 
 import (
 	"io"
-	"os"
-	"syscall"
+	"io/ioutil"
 )
 
 // Output enum of the printer.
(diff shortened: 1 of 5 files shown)


pkg/printer/options.go
```diff
@@ -85,7 +84,7 @@ func Writer(w io.Writer) Option {
 // IgnoreStderr configures the output to not print any
 func IgnoreStderr() Option {
 	return func(opts *Options) {
-		opts.werr = os.NewFile(uintptr(syscall.Stderr), os.DevNull)
+		opts.werr = ioutil.Discard
```
glibsm: neat

PR #192, opened 2020-03-19
filter: Add filters for source and destination identities

michi-covalent:
2 commits:

- implement identity filters.
- add identity filter options to observe command.


michi-covalent:
cc @geakstr @genbit

api/v1/flow/flow.proto
```diff
@@ -252,6 +252,10 @@ message FlowFilter {
     repeated bool reply = 15;
     // dns_query filters L7 DNS flows by query patterns (RE2 regex), e.g. 'kube.*local'.
     repeated string dns_query = 18;
+    // source_identity filters by the security identity of the source endpoint.
+    repeated string source_identity = 19;
```
gandro: Why `string` and not `uint32`? Since it's repeated we can properly differentiate between a missing filter and a zero entry.

michi-covalent: huh good question, i just copied this from `repeated string source_port = 13;` without thinking much. guess this can be integer and not string. let me try changing it.

gandro: Huh, thanks! Why is `source_port` a string though :thinking:?

gandro: I got curious and checked the history. Turns out it's a string since you and me both decided that we might want to support port ranges in the future (e.g. `2000-3000`), thus port is a string.

michi-covalent: k that makes sense. let's use int for identities

michi-covalent: changed to use uint64 to be consistent with https://github.com/cilium/hubble/blob/6e6266b302f016ee50d2f41f70b76f0ecc1d3bb2/api/v1/flow/flow.proto#L114

PR #193, opened 2020-03-19
server: Introduce hooks for extending server functionality

gandro:
This introduces hooks to allow libraries to extend the way Hubble
processes events and flows. An example of this could be the metrics
package, which is currently hard-coded into the decoding pipeline.
This commit introduces three initial hooks in the server:

  1. `OnServerInit`: Invoked after all server options have been parsed.
     This is mostly intended for hooks to obtain the finalized list of
     server properties, for example to obtain a reference to the
     Cilium daemon when running inside Cilium.
  2. `OnMonitorEvent`: Invoked after an Cilium monitor event has been
     observed, but not yet decoded.
  3. `OnDecodedFlow`: Invoked after the parser has decoded the event,
     provides write access to the fully populated `pb.Flow` before it
     is added to the ring buffer.

Co-Authored-By: Glib Smaga <code@gsmaga.com>

---

Actual users of this API (like a revamped `metrics` package) will be added in future PRs, once the feedback (e.g. proposals regarding naming) for this PR has been addressed.

### DIFF
--- a/pkg/server/local_observer.go
+++ b/pkg/server/local_observer.go
@@ -34,6 +34,10 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// DefaultOptions to include in the server. Other packages may extend this
+// in their init() function.
+var DefaultOptions []serveroption.Option
+
 // GRPCServer defines the interface for Hubble gRPC server, extending the
 // auto-generated ObserverServer interface from the protobuf definition.
 type GRPCServer interface {
@@ -87,6 +91,7 @@ func NewLocalServer(
 	options ...serveroption.Option,
 ) (*LocalObserverServer, error) {
 	opts := serveroption.Default // start with defaults
+	options = append(options, DefaultOptions...)
 	for _, opt := range options {
 		if err := opt(&opts); err != nil {
 			return nil, fmt.Errorf("failed to apply option: %v", err)
@@ -108,12 +113,34 @@ func NewLocalServer(
 		opts:          opts,
 	}
 
+	for _, f := range s.opts.OnServerInit {
+		err := f.OnServerInit(s)
+		if err != nil {
+			s.log.WithError(err).Error("failed in OnServerInit")
+			return nil, err
+		}
+	}
+
 	return s, nil
 }
 
 // Start implements GRPCServer.Start.
 func (s *LocalObserverServer) Start() {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+nextEvent:
 	for pl := range s.GetEventsChannel() {
+		for _, f := range s.opts.OnMonitorEvent {
+			stop, err := f.OnMonitorEvent(ctx, pl)
+			if err != nil {
+				s.log.WithError(err).WithField("data", pl.Data).Info("failed in OnMonitorEvent")
+			}
+			if stop {
+				continue nextEvent
+			}
+		}
+
 		flow, err := decodeFlow(s.payloadParser, pl)
 		if err != nil {
 			if !errors.IsErrInvalidType(err) {
@@ -122,7 +149,19 @@ func (s *LocalObserverServer) Start() {
 			continue
 		}
 
+		for _, f := range s.opts.OnDecodedFlow {
+			stop, err := f.OnDecodedFlow(ctx, flow)
+			if err != nil {
+				s.log.WithError(err).WithField("data", pl.Data).Info("failed in OnDecodedFlow")
+			}
+			if stop {
+				continue nextEvent
+			}
+		}
+
+		// FIXME: Convert metrics into an OnDecodedFlow function
 		metrics.ProcessFlow(flow)
+
 		s.GetRingBuffer().Write(&v1.Event{
 			Timestamp: pl.Time,
 			Event:     flow,
@@ -159,6 +198,11 @@ func (s *LocalObserverServer) GetStopped() chan struct{} {
 // GetPayloadParser implements GRPCServer.GetPayloadParser.
 func (s *LocalObserverServer) GetPayloadParser() *parser.Parser {
 	return s.payloadParser
+}
+
+// GetOptions implements serveroptions.Server.GetOptions.
+func (s *LocalObserverServer) GetOptions() serveroption.Options {
+	return s.opts
 }
 
 // ServerStatus should have a comment, apparently. It returns the server status.
--- a/pkg/server/local_observer_test.go
+++ b/pkg/server/local_observer_test.go
@@ -31,7 +31,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestNewLocalServer(t *testing.T) {
+func noopParser(t *testing.T) *parser.Parser {
 	pp, err := parser.New(
 		&testutils.NoopEndpointGetter,
 		&testutils.NoopIdentityGetter,
@@ -40,6 +40,11 @@ func TestNewLocalServer(t *testing.T) {
 		&testutils.NoopServiceGetter,
 	)
 	require.NoError(t, err)
+	return pp
+}
+
+func TestNewLocalServer(t *testing.T) {
+	pp := noopParser(t)
 	s, err := NewLocalServer(pp, logger.GetLogger())
 	require.NoError(t, err)
 	assert.NotNil(t, s.GetStopped())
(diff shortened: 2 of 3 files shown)


pkg/server/serveroption/option.go
```diff
@@ -26,14 +34,45 @@ var Default = Options{
 type Options struct {
 	// Both sizes should really be uint32 but it's better saved for a single
 	// refactor commit.
-
 	MaxFlows      int // max number of flows that can be stored in the ring buffer
 	MonitorBuffer int // buffer size for monitor payload
+
+	CiliumDaemon interface{} // when running inside Cilium, contains a reference to the daemon
```
Rolinh: Does it have to be an empty interface?

pkg/server/local_observer_test.go
```diff
@@ -31,7 +32,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestNewLocalServer(t *testing.T) {
+func noopParser(t *testing.T) *parser.Parser {
```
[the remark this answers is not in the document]
gandro: Unfortunately that would introduce a cyclic dependency between `pkg/parser` and `pkg/testutils`. This function would have to become its own subpackage, e.g. `testutils/parser`. But considering how small the function is and how little it is used, I decided it's not worth creating a new package just for it.

But I'm happy to move it if people feel strongly that this should become its own package.

pkg/server/serveroption/option.go
gandro: The underlying type here will be `github.com/cilium/cilium/daemon/cmd.Daemon`. I opted for an interface instead of the concrete type for two reasons:

- Referencing the concrete type would basically pull in all packages of cilium-agent into `vendor/` (since `Daemon` is the root)
- The concrete as it is defined today is impossible to mock

Unfortunately, at least to me, there is no obvious minimal interface that `Daemon` would satisfy. This type has many ad-hoc methods and getters that make it unpractical to implement a coherent interface. Each hook will want to access a different aspect of Daemon and therefore downcast to a custom interface.

This is what all consumers of `Daemon` inside Cilium already do. To give you three random examples of the various interfaces that `Daemon` implements (there are many more):

- https://github.com/cilium/cilium/blob/cfb7333a064e4bf00ee4bcf95e556b5331b632cd/pkg/datapath/ipcache/listener.go#L49-L51
- https://github.com/cilium/cilium/blob/f0d65d667e33f8a137606015cc37c8d4d3d4e95c/pkg/identity/cache/allocator.go#L106-L116
- https://github.com/cilium/cilium/blob/ce2079869a0385423e802ca22c9233bda31dd93a/pkg/k8s/watchers/watcher.go#L113-L117

Additional examples include the five `pkg/parser/getter` interfaces in Hubble, which are also all references to `Daemon` at runtime.

pkg/server/serveroption/option.go
```diff
@@ -50,3 +89,35 @@ func WithMaxFlows(size int) Option {
...
+	}
+}
+
+// WithAfterDecoding adds a new callback to be invoked after decoding
+func WithAfterDecoding(cb AfterDecoding) Option {
+	return func(o *Options) error {
+		o.AfterDecoding = append(o.AfterDecoding, cb)
+		return nil
+	}
+}
```
glibsm: You're missing the third part of this boilerplate which allows pure functions to act as options. Good example of that is `http.HandlerFunc`. It makes testing easier, and simple usages don't require a conrete type definition to conform to one of the interfaces.

pkg/server/serveroption/option.go
glibsm: Ideally it would be something more than just empty interface. Even it's an interface with one or two methods that is deinfed in *this* pacakge. 

Perhaps something like version? Basic config? Mode? Etc...

pkg/server/local_observer.go
```diff
@@ -34,6 +34,10 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// DefaultOptions to include in the server. Third-party packages may extend this
+// in their init() function.
+var DefaultOptions []serveroption.Option
```
glibsm: In POC this was done as a work-around to not plumb through cilium, I'm not sure this is a great pattern to keep. Are we able to inject the options instead?

pkg/server/serveroption/option.go
gandro: Please feel free to propose a concrete method in that struct. I didn't find any.

glibsm: Maybe start with something as basic as this? https://github.com/cilium/cilium/blob/master/daemon/cmd/daemon.go#L149

pkg/server/local_observer_test.go
```diff
@@ -114,3 +109,71 @@ func TestLocalObserverServer_GetFlows(t *testing.T) {
...
+		pl := &pb.Payload{
+			Time: &types.Timestamp{Seconds: int64(i)},
+			Type: pb.EventType_EventSample,
+			Data: data,
+		}
+		m <- pl
+	}
+	close(s.GetEventsChannel())
+	<-s.GetStopped()
+	assert.Equal(t, h.seenFlows, int64(numFlows))
```
glibsm: assertion is better done in the other direction -- first is the value you expect, the other is the result of running the test.

This will generate a much better error message

pkg/server/local_observer_test.go
```diff
@@ -114,3 +109,71 @@ func TestLocalObserverServer_GetFlows(t *testing.T) {
...
+	return nil
+}
+
+func (h *testHook) BeforeDecoding(ctx context.Context, payload *pb.Payload) (stop bool, err error) {
+	h.seenFlows++
+	assert.Equal(h.t, payload.Time.Seconds, h.seenFlows-1)
+	if h.seenFlows%h.stopEvery == 0 {
+		stop = true
+	}
+	return stop, nil
```
glibsm: naked return here is very confusing, it reads as is this hook should stop processing

pkg/server/local_observer_test.go
```diff
@@ -114,3 +109,71 @@ func TestLocalObserverServer_GetFlows(t *testing.T) {
...
+}
+
+func TestHooks(t *testing.T) {
+	numFlows := 10
+	queueSize := 0
+
+	type fakeCiliumDaemon struct{}
+	h := &testHook{
+		t:            t,
+		stopEvery:    2,
```
glibsm: where and how is this `% 2 == 0` covered in this test?

pkg/server/serveroption/option.go
gandro: Sounds good! I will add it.

pkg/server/local_observer_test.go
gandro: https://github.com/cilium/hubble/blob/bbfe90eb6484a2a462b71e85fb6dc3d6c6311de9/pkg/server/local_observer_test.go#L137-L139

pkg/server/local_observer_test.go
gandro: Fair enough, I'll introduce the early return in the if body. The current is an artifact from an earlier version which had a slightly different control flow.

pkg/server/local_observer_test.go
gandro: Good catch!

pkg/server/serveroption/option.go
gandro: Sure, I can add this back!

pkg/server/local_observer_test.go
glibsm: I see.

I think adding some pure function hooks will make this a lot easier to comprehend as one test https://github.com/cilium/hubble/pull/193/files#r395140700

This also doesn't cover that two hooks A and B both doing `BeforeDecoding` get chained together correctly (first one returning stop).

glibsm:
Once this is merged we can cut 0.5

gandro:
Okay, I believed I have addressed all of your comments, thanks for the feedback!

I have also decided to rename the functions again. I believe the new names make them slightly more self-explanatory with regards on what kind of data they act on and also makes it a bit more clear that these are essentially callbacks.

- `AfterServerInit` → `OnServerInit`
- `BeforeDecoding` → `OnMonitorEvent`
- `AfterDecoding` → `OnDecodedFlow`

 I have also changed the signature of `AfterDecoding` to use a `pb.Flow` instead of `v1.Event` (as the original prototype did). If we end up adding events other than `pb.Flow` we likely want to do the type-based dispatch at the call-site, instead of forcing each hook to implement the same dispatch mechanism over and over again.

pkg/server/serveroption/option.go
```diff
@@ -26,14 +44,63 @@ var Default = Options{
...
+
+	OnServerInit   []OnServerInit   // invoked when the hubble server is initialized
+	OnMonitorEvent []OnMonitorEvent // invoked before an event is decoded
+	OnDecodedFlow  []OnDecodedFlow  // invoked after a flow has been decoded
 }
 
+// returning `stop: true` from a callback stops the execution chain, regardless
+// of the error encountered (for example, explicitly filtering out certain
+// events, or similar).
+type stop = bool
```
Rolinh: I didn't know one could define [type alias](https://golang.org/ref/spec#Type_declarations) instead of just type definitions! Neat and handy.

gandro: The thanks goes to @glibsm for this (he wrote this original code). :)

glibsm: golang 1.9 was eagerly anticipated by library authors for this very reason. This allowed much better backwards compatibility in addition to better API definitions that spelled out what the simple type was used for

PR #200, opened 2020-03-24
Add support for policy verdict events

gandro:
Policy verdict notification events are a new event type added in Cilium
(see cilium/cilium#9943) which are emitted whenever the Cilium datapath
takes a policy decision.

The events are very similar to trace notify events, with some additional
information, i.e. the traffic direction for which a policy verdict has
been taken and the kind of policy match (e.g. L3, L3/L4 or AllowAll).

This PR extends the `threefour` parser with support for policy verdict
notifications. The `seven` parser is extended to also expose the traffic
direction.

Note that policy verdict events are not displayed by the CLI by default.
The user has to manually request them using `-t policy-verdict`.


pkg/parser/seven/parser.go
```diff
@@ -216,6 +218,17 @@ func decodeVerdict(verdict accesslog.FlowVerdict) pb.Verdict {
...
 }
 
+func decodeTrafficDirection(direction accesslog.ObservationPoint) pb.TrafficDirection {
+	switch direction {
+	case accesslog.Ingress:
+		return pb.TrafficDirection_INGRESS
+	case accesslog.Egress:
+		return pb.TrafficDirection_EGRESS
+	default:
+		return pb.TrafficDirection_TRAFFIC_DIRECTION_UNKNOWN
```
glibsm: :O

glibsm:
This PR is rather sparse on tests so I can't get the entire context into my head.

Do these policy verdict notifications contain the full data of the flow, or just very few fields?

gandro:
> Do these policy verdict notifications contain the full data of the flow, or just very few fields?

They contain the full data, they are very similar to trace events. I guess I should extend the parser tests with an example.

pkg/parser/seven/parser.go
gandro: The naming is not great, but that's due to protobuf not having namespaced enums. This requres us to prefix `UNKNOWN`s at the protobuf level, otherwise it would collide with other enums that have an `UNKNOWN` variant :man_shrugging:

glibsm:
> They contain the full data, they are very similar to trace events. I guess I should extend the parser tests with an example.

How _very_ similar. Are we talking duplicate data level similar in the output? Or these completely obsolete trace when -t is used.

gandro:
> How very similar. Are we talking duplicate data level similar in the output? Or these completely obsolete trace when -t is used.

They are similar in the sense that they contain the first few bytes of the packet payload and contain metadata like remote numerical identity. But I wouldn't say they replace the trace event, rather they complement them. Trace events help to put a policy verdict event in context.

In this regard, trace events are still important, because they tell us what happened before and after a policy verdict event was reached. For example, the trace notify `from-endpoint` event will give us access to the packet before service translation, and trace events can tell us whether a packet was forwarded to the network or directly delivered to a local endpoint.

It's not the perfect comparison, but think of policy verdict events being the dual to drop events, as even a drop event is always preceded by some trace events.

pkg/printer/printer.go
```diff
@@ -168,10 +166,28 @@ func GetFlowType(f v1.Flow) string {
...
 	}
-	if f.GetVerdict() == pb.Verdict_DROPPED {
-		return api.DropReason(uint8(f.GetEventType().SubType))
+
+	switch f.GetEventType().GetType() {
+	case api.MessageTypeTrace:
+		return api.TraceObservationPoint(uint8(f.GetEventType().GetSubType()))
+	case api.MessageTypeDrop:
+		return api.DropReason(uint8(f.GetEventType().GetSubType()))
+	case api.MessageTypePolicyVerdict:
```
michi-covalent: would be nice if we can somehow reuse the logic from https://github.com/cilium/cilium/blob/e831859b5cc336c6d964a6d35bbd34d1840e21b9/pkg/monitor/datapath_policy.go#L104

pkg/parser/threefour/parser.go
```diff
@@ -305,6 +315,8 @@ func decodeVerdict(eventType uint8) pb.Verdict {
 		return pb.Verdict_DROPPED
 	case monitorAPI.MessageTypeTrace:
 		return pb.Verdict_FORWARDED
+	case monitorAPI.MessageTypePolicyVerdict:
+		return pb.Verdict_FORWARDED
```
michi-covalent: trying to understand this part. so the verdict field of the verdict message is always forwarded? :)

api/v1/flow/flow.proto
```diff
@@ -64,6 +64,13 @@ message Flow {
     // destination_service contains the service name of the destination
     Service destination_service = 21;
 
+    // traffic_direction of the connection, e.g. ingress or egress
+    TrafficDirection traffic_direction = 22;
+
+    // policy_match_type is only applicable to the cilium event type PolicyVerdict
+    // https://github.com/cilium/cilium/blob/e831859b5cc336c6d964a6d35bbd34d1840e21b9/pkg/monitor/datapath_policy.go#L50
+    uint32 policy_match_type = 23;
```
michi-covalent: possibly this field needs to be an enum? otherwise the consumers of this API need to know what all the integers mean. not a blocker though.

pkg/printer/printer.go
gandro: Yes, unfortunately that function is not exported. I can open a PR on Cilium to make it public and undo the logic, but I don't think this should block this PR.

api/v1/flow/flow.proto
gandro: I choose u32 to match `drop_reason`, which is basically it's analog for dropped flows. In earlier versions I even called the field `forward_reason`, but that doesn't align with `flow.Verdict` (which is FORWARDED even for trace events which don't have a policy match)

pkg/printer/printer.go
michi-covalent: no not a blocker

pkg/parser/threefour/parser.go
gandro: Yes. The naming might be a bit confusing. But a verdict event is only ever emitted if there is a policy that matches the flow. If there is not policy event, then the flow will be dropped.

gandro:
> does this mean double the number of flows, or do these get filtered out by default by cilium event filter?

For gRPC requests from the CLI they get filtered out by default for now, i.e. they are not part of the default event types.

Also, luckily policy verdict events are much less frequent than trace events :)

A policy verdict is generated for each _connection_ at most twice (once ingress, once egress). 

Compare this to the trace notify events which are generated for IP _packet_ on each TCP state change (assuming monitor aggregation is set to medium) at multiple places in the datapath for the same packet (`from-endpoint`, `to-stack`, `from-stack`, `to-endpont` etc). So for each policy verdict event, there are typically at least 6-8 trace notify events.

michi-covalent:
sounds good to me

api/v1/flow/flow.proto
gandro: Oh, I have to correct myself. A policy verdict event is also emitted when the flow is about to be dropped. I will fix this!

gandro:
Alright, thanks for the feedback everyone. I've pushed a new version which contains some unit tests and handles policy verdict events for dropped flows properly (see https://github.com/cilium/cilium/blob/aea00a0e38dfbf85714cfdd905525bcc988be6b7/bpf/bpf_lxc.c#L170-L178). Thanks for @michi-covalent for pointing that out, this is something I was not aware of.

pkg/parser/threefour/parser.go
gandro: I was wrong. Adapted the code to return the proper verdict.

PR #206, opened 2020-03-26
api: Small fixes to the protoc invocations in Makefile

gandro:
Our `api/v1` Makefile invokes the `protoc-gen-validate` plugin. We should vendor it in to avoid calling into the system-wide installation, as the latter might differ on individual development setups.

In addition, this PR also removes the `$PATH` hack in the Makefile and instead relies on the `--plugin` argument of protoc to pull in our vendored plugins.

Also bumps the required `protoc` version to 3.11.4. It's the newest version of protoc. It still produces the same generated Go code for Hubble, so the generated files are not affected.


go.mod
```diff
@@ -4,6 +4,7 @@ go 1.14
 
 require (
 	github.com/cilium/cilium v1.7.0-rc2.0.20200311180626-711b37ed100c
+	github.com/envoyproxy/protoc-gen-validate v0.3.0-java
```
Rolinh: ¯\_(ツ)_/¯

gandro: For the record, this is auto-generated by `go mod`. I believe this is the `minimal version` selected by `go mod` due to a transitive dependency we have on envoyproxy via Cilium.

PR #209, opened 2020-04-01
OnBuildFilter

tgraf:


pkg/server/local_observer.go
```diff
@@ -394,7 +395,7 @@ func (r *flowsReader) Next(ctx context.Context) (*pb.Flow, error) {
 
 // newRingReader creates a new RingReader that starts at the correct ring
 // offset to match the flow request.
-func newRingReader(ring *container.Ring, req *observer.GetFlowsRequest) (*container.RingReader, error) {
+func newRingReader(ctx context.Context, ring *container.Ring, req *observer.GetFlowsRequest) (*container.RingReader, error) {
```
Rolinh, line 398: I'm not fond of the idea of passing a context to a constructor. Why does `BuildFilter` require a context in the first place? I mean, I noticed that `OnBuildFilter` requires a context but I don't understand why. Is building filters an operation that is expected to be running for a very long time and that needs to be cancelable?

pkg/server/local_observer.go
```diff
@@ -315,12 +315,13 @@ type flowsReader struct {
 // newFlowsReader creates a new flowsReader that uses the given RingReader to
 // read through the ring buffer. Only flows that match the request criterias
 // are returned.
-func newFlowsReader(r *container.RingReader, req *observer.GetFlowsRequest, log *logrus.Entry) (*flowsReader, error) {
-	whitelist, err := filters.BuildFilterList(req.Whitelist)
+func (s *LocalObserverServer) newFlowsReader(ctx context.Context, r *container.RingReader, req *observer.GetFlowsRequest, log *logrus.Entry) (*flowsReader, error) {
```
Rolinh, line 318: Same comment as for `newRingReader` about the context passed to this constructor. I'm also a bit concerned about this constructor becoming a `LocalObserverServer` method. Isn't it possible to avoid this?

pkg/server/local_observer.go
```diff
@@ -220,10 +220,10 @@ func (s *LocalObserverServer) GetFlows(
 	req *observer.GetFlowsRequest,
 	server observer.Observer_GetFlowsServer,
 ) (err error) {
-	return getFlows(req, server, s)
+	return s.getFlows(req, server, s)
 }
 
-func getFlows(
+func (s *LocalObserverServer) getFlows(
```
Rolinh, line 226: Why does this function need to become a `LocalObserverServer` method?

tgraf:
> I wonder though if we should use a handler func instead of `struct{}` types for the implementors, which is what we have for the other hooks.

I've chosen this path primarily to be future proof so we can add additional functions to these filter plugins as needed.

michi-covalent:
lgtm overall! a couple of things:

- as @Rolinh said, not sure if we need to pass context to build filters.
- how does this interact with json export? right now it calls `filters.BuildFilterList` directly: https://github.com/isovalent/atlantis-plugin-export/blob/fd910be1684f25fc88d9188aeee1a23fcca4ca08/processor.go#L44

pkg/server/local_observer.go
tgraf: Because access to `s.opts` is required to retrieve the list of `OnBuildFilter`:

```
        filterList := append(filters.DefaultFilters, s.opts.OnBuildFilter...)
```

pkg/server/local_observer.go
tgraf: I don't really understand the constructor comment. The GetFlows function handles a request and the context is passed as the request is being handled. `newRingReader()` is part of this, it only exists for the duration of the API request and the context is required in the `OnBuildFilter()`.

tgraf: Fixed as discussed on Slack

pkg/server/local_observer.go
tgraf: Fixed as discussed on Slack

pkg/server/local_observer.go
tgraf: Restructured so this receiver is no longer required.

pkg/filters/event_type.go
```diff
@@ -0,0 +1,58 @@
...
+type EventTypeFilter struct{}
+
+// OnBuildFilter builds an event type filter
+func (e *EventTypeFilter) OnBuildFilter(ctx context.Context, ff *pb.FlowFilter) ([]FilterFunc, error) {
+	types := ff.GetEventType()
+	if len(types) > 0 {
+		return []FilterFunc{filterByEventType(types)}, nil
+	}
+
+	return nil, nil
```
glibsm, line 57: `nil, nil` is a dangerous return to handle correctly as people are only used to checking `!= nil` on error

tgraf: I had an empty slice in the beginning but then optimized it. I'll revert it back.

PR #212, opened 2020-04-03
add peer gRPC service

Rolinh:
This service is required for peer discovery in the context of the multi-node feature. It will be implemented by embedded Hubble using Cilium's node manager and be used by Hubble proxy to discover Hubble peers and get notified of changes.

I've implemented the gRPC service locally to see if it fits the needs and I can confirm that it does.
It initially consisted of 2 rpc calls but after a discussion with @gandro, we concluded that trimming down to a single rpc call would not only simplify implementation and usage but would also prevent a class of bugs where races could occur (between a call to list and notify for example) leading a client to eventually miss peer change notifications.

Ref: #89


api/v1/peer/peer.proto
```diff
@@ -0,0 +1,50 @@
...
+
+package peer;
+
+// Peer lists  hubble peers and notifies of changes.
+service Peer {
+    // Notify sends information about hubble peers in the cluster.
+    // When Notify is called, it sends information about all the peers that are
+    // already part of the cluster (with the type as PEER_ADDED). It
+    // subsequently notifies of any change.
+    rpc Notify(google.protobuf.Empty) returns (stream ChangeNotification) {}
```
gandro, line 27: I think we might want to define an empty `message NotifyRequest` here just in case, similar to `StatusRequest` on `service Observer`.
https://github.com/cilium/hubble/blob/9bd43f6deadfc4c80b2593573e40c662f9422e12/api/v1/observer/observer.proto#L32

It allows us to add optional request parameters in the future if we ever need it.

api/v1/peer/peer.proto
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// ChangeNotification indicates a change regarding a hubble peer.
+message ChangeNotification {
+    // Name is the name of the peer, typically the hostname. The name includes
+    // the cluster name if a value other than default has been specified.
```
gandro, line 33: nit: Maybe clarify that this uniquely identifies a host. An example would also be nice (i.e. how does it include the cluster name).

PR #213, opened 2020-04-04
server: Match time range before filters

tgraf:
Filters are potentially expensive, time ranges have a constant cost.
Match it first.

Signed-off-by: Thomas Graf <thomas@cilium.io>

### DIFF
--- a/pkg/server/local_observer.go
+++ b/pkg/server/local_observer.go
@@ -375,20 +375,20 @@ func (r *flowsReader) Next(ctx context.Context) (*pb.Flow, error) {
 		if e == nil {
 			return nil, io.EOF
 		}
+		if r.timeRange {
+			ts, err := types.TimestampFromProto(e.GetFlow().GetTime())
+			if err != nil {
+				return nil, err
+			}
+			if ts.After(r.end) {
+				return nil, io.EOF
+			}
+			if ts.Before(r.start) {
+				continue
+			}
+		}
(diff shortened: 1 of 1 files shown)


pkg/server/local_observer.go
```diff
@@ -375,20 +375,20 @@ func (r *flowsReader) Next(ctx context.Context) (*pb.Flow, error) {
 		if e == nil {
 			return nil, io.EOF
 		}
+		if r.timeRange {
+			ts, err := types.TimestampFromProto(e.GetFlow().GetTime())
```
Rolinh, line 379: Now that this has been moved before the type cast, `e.Getflow()` might return `<nil>`. So this code might potentially lead to a panic triggered by a nil pointer dereference.
What about assigning and checking the flow validity once just above after the `e == nil` check above?

gandro: `v1.Event.GetFlow()` returns a typed `nil` and `pb.Flow.GetTime` is able to deal with the receiver being `nil`. `types.TimestampFromProto` also treats nil as the `zero` time, so it should not cause a panic. 

But I agree that introducing a nil check might be good practice either way however, in case any of the implementations change.

Rolinh: Ha, fair enough. Let's leave that up to @tgraf, shouldn't block the PR then.

glibsm:
Can follow up fix later if needed

PR #214, opened 2020-04-04
Add OnGetFlows and OnFlowDelivery

tgraf:
Depends on #213

### DIFF
--- a/pkg/server/local_observer.go
+++ b/pkg/server/local_observer.go
@@ -257,6 +257,7 @@ func (s *LocalObserverServer) GetFlows(
 		return err
 	}
 
+nextFlow:
 	for ; ; i++ {
 		flow, err := flowsReader.Next(server.Context())
 		if err != nil {
@@ -265,6 +266,17 @@ func (s *LocalObserverServer) GetFlows(
 			}
 			return err
 		}
+
+		for _, f := range s.opts.OnFlowDelivery {
+			stop, err := f.OnDecodedFlow(server.Context(), flow)
+			if err != nil {
+				return err
+			}
+			if stop {
+				continue nextFlow
+			}
+		}
+
 		err = server.Send(&observer.GetFlowsResponse{
 			ResponseTypes: &observer.GetFlowsResponse_Flow{
 				Flow: flow,
--- a/pkg/server/local_observer.go
+++ b/pkg/server/local_observer.go
@@ -230,18 +230,9 @@ func (s *LocalObserverServer) GetFlows(
 		return err
 	}
 
-	return getFlows(req, server, s, whitelist, blacklist)
-}
-
-func getFlows(
-	req *observer.GetFlowsRequest,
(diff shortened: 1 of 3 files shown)


pkg/server/local_observer.go
```diff
@@ -220,12 +220,20 @@ func (s *LocalObserverServer) GetFlows(
 	req *observer.GetFlowsRequest,
 	server observer.Observer_GetFlowsServer,
 ) (err error) {
+	ctx := server.Context()
```
gandro, line 223: I think we should actually wrap this context here in a `WithCancel` and call cancel when `GetFlows` returns, to ensure that the context lives only as as long as the request itself. This allows implementors of the `OnGetFlows` hook to release any resources it allocated for this particular request.

We probably also want to pass this per-request-context down to the `OnBuildFilter` and `OnFlowDelivery` hooks as well.

pkg/server/local_observer.go
```diff
@@ -259,7 +267,7 @@ func (s *LocalObserverServer) GetFlows(
 
 nextFlow:
 	for ; ; i++ {
-		flow, err := flowsReader.Next(server.Context())
+		flow, err := flowsReader.Next(ctx)
```
gandro, line 270: While not introduced by this PR, I just notice it now: Passing down `server.Context()` to the flow reader is wrong. I believe this is another reason we want introduce a per-request context (as stated above).

To demonstrate: The flow reader spawns goroutines to implement `follow` requests. These goroutines must be terminated once the request is finished, which is achieved via context cancellation. If we pass down `server.Context()` here, these per-request goroutines will be kept alive until the server is terminated, which is essentially a goroutine leak.  The context passed down to `flowsReader.Next` must only live as long as the request.

pkg/server/local_observer_test.go
```diff
@@ -231,3 +231,64 @@ func TestLocalObserverServer_OnFlowDelivery(t *testing.T) {
...
+		FakeGRPCServerStream: &FakeGRPCServerStream{
+			OnContext: func() context.Context {
+				return context.Background()
+			},
+		},
+	}
+
+	var key int
+	onGetFlows := func(ctx context.Context, req *observer.GetFlowsRequest) (context.Context, error) {
+		return context.WithValue(ctx, key, 10), nil
```
gandro, line 254: See `golint`: `should not use basic type int as key in context.WithValue`

Introducing a `type ctxKey int` should be sufficient to satisfy the lint.

pkg/server/local_observer.go
gandro: Update: I have fixed the context in #216.

pkg/server/local_observer_test.go
tgraf: Yep, this will need a type we can use for this consistently.

tgraf:
Closing in favor of https://github.com/cilium/cilium/pull/10896

Issue #215, opened 2020-04-06
goroutine leak in GetFlows request


gandro:
We are passing down `server.Context()` to the flow reader here:
https://github.com/cilium/hubble/blob/bbfe90eb6484a2a462b71e85fb6dc3d6c6311de9/pkg/server/local_observer.go#L260

This however is wrong. The flow reader spawns goroutines to implement `follow` requests. These goroutines must be terminated once the request is finished, which is achieved via context cancellation:

https://github.com/cilium/hubble/blob/9bd43f6deadfc4c80b2593573e40c662f9422e12/pkg/container/ring_reader.go#L65-L67

If we pass down `server.Context()` to `flowReader.Next()`, these per-request goroutines will be kept alive until the server is terminated, which is essentially a goroutine leak.  The context passed down to `flowsReader.Next` must only live as long as the request. We can fix this by wrapping `server.Context()` with `WithCancel` and cancel the wrapped context when `GetFlows` is about to return.

PR #216, opened 2020-04-06
server: Introduce per-request context

gandro:
The ring ringer spawns goroutines when serving a `follow` request.
These goroutines must terminate when the request is done, otherwise
the goroutines are leaked. Therefore this commit introduces a
per-request context for the `GetFlows` gRPC call which is cancelled
when the request is completed, allowing consumers of the context to
terminate as well.

Fixes: #215

### DIFF
--- a/pkg/server/local_observer.go
+++ b/pkg/server/local_observer.go
@@ -220,20 +220,27 @@ func (s *LocalObserverServer) GetFlows(
 	req *observer.GetFlowsRequest,
 	server observer.Observer_GetFlowsServer,
 ) (err error) {
+	// This context is used for goroutines spawned specifically to serve this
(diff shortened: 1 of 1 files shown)


pkg/server/local_observer.go
```diff
@@ -220,20 +220,27 @@ func (s *LocalObserverServer) GetFlows(
 	req *observer.GetFlowsRequest,
 	server observer.Observer_GetFlowsServer,
 ) (err error) {
+	// This context is used for goroutines spawned specifically to serve this
+	// request, meaning it must be cancelled once the request is done and this
+	// function returns.
+	ctx, cancel := context.WithCancel(server.Context())
```
glibsm, line 226: http has context as part of the request but grpc doesn't 🤔

PR #217, opened 2020-04-06
cmd/serve: remove global default options

Rolinh:
This commits removes the global variable DefaultOptions in favor of a
private constructor.

The problem with DefaultOptions as a global variable is that it gets
modified the first time NewServer is called. Thus, subsequent calls to
NewServer instantiate a server with modified default options (with an
existing listener typically) leading to unpredictable behavior.

### DIFF
--- a/cmd/serve/option.go
+++ b/cmd/serve/option.go
@@ -29,17 +29,19 @@ import (
 	healthpb "google.golang.org/grpc/health/grpc_health_v1"
 )
 
-// DefaultOptions is the reference point for default values.
-var DefaultOptions = Options{
-	Listeners: make(map[string]net.Listener),
-}
-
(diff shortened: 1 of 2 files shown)


glibsm:
When are we ever creating more than one server in the process life of a binary?

Rolinh:
> When are we ever creating more than one server in the process life of a binary?

In the context of embedded hubble for the multi-node feature (this is while working on this that I found out about this issue actually). There's one server that implements the peer+observer gRPC services and that listens on a local unix domain socket and another one that listens on a given set of addresses and that only exposes the observer gRPC service.

gandro:
@Rolinh I believe this PR is obsolete now, correct?

Rolinh:
@gandro Correct. A fix is still required but has to be done in the cilium repo.

PR #218, opened 2020-04-07
cmd: Introduce hubble_{serve,pprof} build tags

gandro:
This introduces a new build tag for both the standalone hubble serve
command, as well as the profiling intrastructure. Both features are
enabled in our Makefile by default, so this should not introduce any
functional changes to current end-users.

The motivation for this change is to enable a hubble/cilium hybrid
client binary that depending on its `argv[0]` name will either dispatch
to the Hubble or Cilium client command-line interface. Since we do not
want to ship the Hubble standalone server code in a binary intended for
client use only, the serve subcommand (and pprof) is only included when
the corresponding `hubble_serve` build tag is set.

Profiling is automatically enabled if `hubble_serve` is set, but can
also be enabled separately for client use with `hubble_pprof`.


gandro:
Made obsolete by #220

PR #220, opened 2020-04-08
Remove all server-side code

tgraf:
After the successful merging of https://github.com/cilium/cilium/pull/10860, all server side code now lives in cilium/cilium and can be removed from this repository.


cmd/root.go
```diff
@@ -114,12 +111,8 @@ func init() {
 
 	rootCmd.SetVersionTemplate("{{with .Name}}{{printf \"%s \" .}}{{end}}{{printf \"v%s\" .Version}}\n")
 
-	l := logger.GetLogger()
-
 	// initialize all subcommands
 	rootCmd.AddCommand(observe.New())
-	rootCmd.AddCommand(serve.New(l))
```
gandro, line 121: This PR completely removes our standalone server required to run Hubble with Cilium 1.6 and 1.7. It this intended?

tgraf: Yes, we have the 0.5.0 release out there to support the standalone mode.

tgraf: I've added a commit to ping images to the 0.5.0 release

Rolinh: Does this mean we have to branch out v0.5 for bugfixes, etc? It seems more work than just keep the `serve` command in the tree for now (maybe with a deprecation notice and/or marking it hidden).

go.mod
```diff
@@ -3,29 +3,16 @@ module github.com/cilium/hubble
...
 	google.golang.org/grpc v1.26.0
-	k8s.io/apimachinery v0.18.0
 )
 
-replace (
-	github.com/miekg/dns => github.com/cilium/dns v1.1.4-0.20190417235132-8e25ec9a0ff3
-	github.com/optiopay/kafka => github.com/cilium/kafka v0.0.0-20180809090225-01ce283b732b
-	k8s.io/client-go => github.com/cilium/client-go v0.0.0-20200326103132-fe7bd31c2794
-)
+replace github.com/optiopay/kafka => github.com/cilium/kafka v0.0.0-20180809090225-01ce283b732b
```
Rolinh, line 18: The replace directives should be identical to the one of the cilium version that is imported (ie: they should remain as they were) or we risk unintended side-effects.

cmd/root.go
gandro: Ah okay, yeah I was also under the impression we would keep support in the `master` tree, but having a `v0.5` branch works for me as well. 

It will signal very clearly that we are not supporting standalone mode anymore besides urgent fixes. But yes, it does mean that backports to v0.5 will involve a lot of manual labor.

(sidenote: if we want to keep support in master, PR #218 allows us to enable/disable the `serve` command at build-time)

cmd/root.go
```diff
@@ -114,12 +111,8 @@ func init() {
...
-	rootCmd.AddCommand(status.New())
```
gandro, line 122: I think there can still be a case to me made for the status command to stick around. It is only accessing the gRPC API, so it works nicely as a connectivity check from client to server. I see this being useful to app developers who want to troubleshoot their K8s application. 

The Hubble section to be added in `cilium status` on the other side I see more useful for infra admins deploying Cilium/Hubble. It also uses a different method to connect to the server: `cilium status` is using the local cilium unix socket only available on the node running cilium-agent, while `hubble status` would be intented to work from a client laptop.

cmd/root.go
tgraf: Yes, we will have to branch out if we want to push bugfixes into 0.5.0. We do the same with Cilium.

If we keep the code around, somebody will make changes.

cmd/root.go
tgraf: Makes sense. We can keep it.

cmd/root.go
Rolinh: Would it be possible then to push this serve command removal to a separate commit so that we can branch out from the server-side code removal commit in order to pick up fixes that have already been committed to master since v0.5 was tagged?

tgraf: Are you asking to create a commit which changes the Hubble to use the server-side code in cilium/cilium first before removing all of that code?

Rolinh: Yes. I think it would make it easier to pickup fixes for hubble server that are made in cilium's repository. We'll have to backport them manually otherwise.

tgraf: Are you OK to do this work once we need to push a fix to the 0.5.0 branch? We can do such a change as the first commit if we ever have to branch off. It's also a bit unclear how that would look like from a versioning perspective. The server side code in cilium is only in master, unreleased, would we have the 0.5 Hubble branch point to an arbitrary Cilium commit in unreleased master?

Rolinh: > Are you OK to do this work once we need to push a fix to the 0.5.0 branch? 

Yes. I just want to make sure that we keep the overhead of maintaining the 0.5 branch at a minimum.

PR #224, opened 2020-04-09
defaults: Introduce new defaults for embedded Hubble

gandro:
Cilium-embedded Hubble uses a different default socket path than the
removed standalone server. This commit introduces a new defaults package
(to reduce the number of dependencies we have on `github.com/cilium/cilium`)
with the new socket path.

In addition, it also unifies and adjusts the various dial and request
default timeouts in `hubble observe` and `hubble status`.


gandro:
Merging this to unblock another PR I want to make.

PR #229, opened 2020-04-10
Missing " character at the end of the container's image key

k3a:


Rolinh:
Thanks. This is a duplicate of #225 though.

PR #234, opened 2020-04-15
actions: Trigger on release branches

michi-covalent:


michi-covalent:
use https://github.com/cilium/hubble/pull/235 instead

PR #240, opened 2020-04-22
Clean up Helm templates

michi-covalent:
- Remove all the existing templates since the Hubble server now runs as
  a part of Cilium daemonset.
- Add a new template to deploy Hubble CLI as a daemonset.
- Add a pre-generated YAML to tutorials/explore-flow-queries.


Rolinh:
> Add a new template to deploy Hubble CLI as a daemonset.

I thought we would only do this for the CI? Hubble-relay is right around the corner.

gandro:
Nice! So I'm starting to think that we will want to keep maintaining a DaemonSet for CI, logging and debugging purposes, even once `hubble-relay` lands. These are use-cases where we want to connect directly to embedded Hubble, without going through relay, and for that we need a Daemon Set.

However, we've just landed a Hubble DaemonSet as part of the Cilium charts: https://github.com/cilium/cilium/pull/11084 - so I'm not sure if it makes a lot of sense to have a similar DaemonSet in this repository here as well.

In addition, before we remove the standalone charts from the `master` branch here, we need to update the Cilium 1.7 docs (which still refer to the master branch) first. I will open a PR for that on Cilium.

michi-covalent:
looks like i can just use https://github.com/cilium/cilium/pull/11084/files#diff-0677ef214d150d971f719f32d18a71b1

i'll close this pr, and open a separate pr to add quickstart.md.

PR #241, opened 2020-04-23
observe: Show all the event types by default

michi-covalent:
Filtering out certain event types by default on the client-side does
not make sense any more since the server already ignores events like
MessageType{Agent,Capture,Debug}. It can also be confusing when new
event types get added since they don't show up if you simply run
`hubble observe` without explicitly whitelisting the new event type.


michi-covalent:
cc @gandro

PR #248, opened 2020-04-28
cmd: add hidden 'peer' command

Rolinh:
This command shows information about Hubble peers by querying the peer gRPC service. It's probably only useful for developers and/or to debug something with hubble-relay (which is why it's hidden).

Note that the peer service is NOT implemented by hubble-relay and is only available when querying hubble server via a local Unix domain socket.

Example usage:

    $ hubble peer watch
    PEER_ADDED   k8s1 192.168.34.11
    PEER_ADDED   k8s2 192.168.34.12
    PEER_ADDED   k8s3 192.168.34.13
    PEER_UPDATED k8s2 192.168.34.12
    ...

Note: this is a command I originally implemented to experiment with the peer gRPC service and then I thought it could be useful so created this PR. Feel free to just close it if you think it's not.


cmd/peer/peer.go
```diff
@@ -0,0 +1,105 @@
...
+var (
+	serverURL     string
+	serverTimeout time.Duration
+)
+
+// New creates a new hidden peer command.
+func New() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "peer",
+		Aliases: []string{"peers"},
```
glibsm, line 40: the reverse reads better: peers by default, peer alias

`hubble peers list` rather than `hubble peer list` and would generate better help page probably

cmd/peer/peer.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to dial grpc: %v", err)
+	}
+	return conn, nil
+}
+
+func newWatchCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:     "watch",
+		Aliases: []string{"monitor"},
```
glibsm, line 70: alias `w` is better

cmd/peer/peer.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+// New creates a new hidden peer command.
+func New() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "peer",
+		Aliases: []string{"peers"},
+		Short:   "Get information about Hubble peers",
+		Long:    `Get information about Hubble peers.`,
+		Hidden:  true, // this command is only useful for development/debugging purposes
```
glibsm, line 43: Is it though? I'm not sure we should be hiding it from `--help` even if it's only useful for debugging

Rolinh: At the moment, the peer service is only available through a local unix domain socket and is not served by hubble-relay. Thus, I think it might confuse users if they have their hubble cli pointing to hubble-relay and some commands don't work which is why I thought that it'd be better to hide this command (at least for now).

cmd/peer/peer.go
Rolinh: It does. Updated.

cmd/peer/peer.go
Rolinh: Done

PR #250, opened 2020-04-29
Add reflect command

michi-covalent:
not cleaned up at all at this point, but i'm posting it here so that i don't lose it.


cmd/reflect/reflect.go
```diff
@@ -0,0 +1,194 @@
...
+func New() *cobra.Command {
+	reflectCmd := &cobra.Command{
+		Use:   "reflect",
+		Short: "Let Hubble reflect upon itself",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runReflect()
+		},
+	}
+	reflectCmd.Flags().StringVarP(&serverURL, "server", "", defaults.GetDefaultSocketPath(), "URL to connect to server")
+	reflectCmd.Flags().StringVarP(&output, "output", "o", "compact", "\"compact\" or \"json\"")
```
glibsm, line 47: `"compact" or "json"` avoids escaping

cmd/reflect/reflect.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+func runReflect() (err error) {
+	conn, err := grpc.Dial(serverURL, grpc.WithInsecure())
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if closeErr := conn.Close(); closeErr != nil {
+			err = closeErr
```
Rolinh, line 58: I think it would be better to assign `err` only if it's `nil` because you will otherwise overwrite the original error. We could also consider error wrapping.

michi-covalent: ah yes that makes sense. i just put some random stuff i found on stackoverflow, trying to figure out what the best practice is for dealing with errors returned from deferred functions.

michi-covalent:
let's use https://github.com/cilium/hubble/pull/378 instead

PR #257, opened 2020-05-11
helm: ensure we can override the priority class name 

mvisonneau:
When running in a namespace other than kube-system, we currently cannot override the `priorityClassName`

### DIFF
--- a/install/kubernetes/hubble/values.yaml
+++ b/install/kubernetes/hubble/values.yaml
@@ -67,9 +67,10 @@ ui:
   #         cpu: 100m
   #         memory: 64Mi
   resources: {}
+  
   # The priority class system-node-critical marks add-on pods as critical to the node itself.
   # This priority class is only valid under the kube-system namespace.
-  priorityClassName: ""
+  priorityClassName: system-node-critical
 
 ingress:
   enabled: false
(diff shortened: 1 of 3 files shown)


install/kubernetes/hubble/values.yaml
```diff
@@ -67,9 +67,10 @@ ui:
   #         cpu: 100m
   #         memory: 64Mi
   resources: {}
+  
   # The priority class system-node-critical marks add-on pods as critical to the node itself.
   # This priority class is only valid under the kube-system namespace.
-  priorityClassName: ""
+  priorityClassName: system-node-critical
```
Rolinh, line 73: This priority class is only available in the `kube-system` namespace so with the namespace name check removed, I think this is going to fail.

Rolinh:
Thanks for your contribution @mvisonneau ! See comment (these charts will be removed from this repository soon).

install/kubernetes/hubble/values.yaml
mvisonneau: indeed, my bad this was not intended 🤦

gandro:
Closing this, as these charts have moved to the Cilium repository.

PR #260, opened 2020-05-22
printer: Add support for NodeStatusEvent

gandro:
This adds support for the new `NodeStatusEvent` (emitted by hubble-relay) message in the printer. 

Currently only prints important node status events, such as `UNAVAILABLE` and `ERROR`. All other events are hidden by default and can be shown by enabling debug output.

Example output:
```console
$ ./hubble observe --server 192.168.33.11:4245 --ip foo -o table
May 25 11:27:33.707 [k8s1]: Error "invalid IP address in filter: \"foo\"" on 2 nodes: k8s2, k8s1
$ ./hubble observe --server 192.168.33.11:4245 --ip foo -o compact
May 25 11:27:36.419 [k8s1]: Error "invalid IP address in filter: \"foo\"" on 2 nodes: k8s1, k8s2
$ ./hubble observe --server 192.168.33.11:4245 --ip foo -o dict   
  TIMESTAMP: May 25 11:27:39.717
      STATE: NODE_ERROR
      NODES: k8s2, k8s1
    MESSAGE: "invalid IP address in filter: \"foo\""
$ ./hubble observe --server 192.168.33.11:4245 --ip foo -o json 2>&1 | jq
{
  "node_status": {
    "state_change": "NODE_ERROR",
    "node_names": [
      "k8s2",
      "k8s1"
    ],
    "message": "invalid IP address in filter: \"foo\""
  },
  "node_name": "k8s1",
  "time": "2020-05-25T11:27:53.868333995Z"
}

```

See https://github.com/cilium/cilium/pull/11589/ for the server-side implementation.


pkg/printer/printer.go
```diff
@@ -78,6 +80,8 @@ func New(fopts ...Option) *Printer {
 const (
 	tab     = "\t"
 	newline = "\n"
+
+	nodeStatusCutoff = 10
```
Rolinh, line 84: I think that `10` is already too many as nodes typically have long names (eg `ip-192-168-12-253.us-west-2.compute.internal`) in practice. What about `2` or `3` instead? Or even better: display as many names as you can given a fixed string length.

gandro:
I have added support for dict and json formatting. This should not break any existing JSON consumers, since the output is printed on stderr.

cmd/observe/observe.go
```diff
@@ -408,12 +414,11 @@ func getFlows(client observer.ObserverClient, req *observer.GetFlowsRequest) err
...
 		}
 
-		flow := getFlowResponse.GetFlow()
-		if flow == nil {
-			continue
+		if flow := getFlowResponse.GetFlow(); flow != nil {
+			err = printer.WriteProtoFlow(flow)
+		} else if status := getFlowResponse.GetNodeStatus(); status != nil {
+			err = printer.WriteProtoNodeStatusEvent(getFlowResponse)
 		}
```
glibsm, line 421: else debug "unknown type" maybe?

pkg/printer/printer.go
```diff
@@ -265,6 +267,94 @@ func (p *Printer) WriteProtoFlow(f v1.Flow) error {
 	return nil
 }
 
+func joinWithCutOff(elems []string, sep string, targetLen int) string {
```
glibsm, line 270: comment the function, please

pkg/printer/printer_test.go
```diff
@@ -373,29 +373,32 @@ func Test_getTimestamp(t *testing.T) {
...
 			},
 			want: "Jan  1 00:00:00.000",
 		},
 		{
 			name: "invalid",
-			args: args{},
+			args: args{
+				t: &timestamp.Timestamp{
+					Seconds: 0,
+					Nanos:   2_000_000_000,
```
glibsm, line 388: casting `X * time.Seconds` would probably be even more legible here

gandro: It's supposed to be an invalid timestamp. So casting doesn't work. I used negative numbers instead now, to make it more legible that it is supposed to be a garbage value.

pkg/printer/printer.go
```diff
@@ -265,6 +267,99 @@ func (p *Printer) WriteProtoFlow(f v1.Flow) error {
...
+		_, err := fmt.Fprint(p.opts.werr,
+			"  TIMESTAMP: ", fmtTimestamp(r.GetTime()), newline,
+			"      STATE: ", s.StateChange.String(), newline,
+			"      NODES: ", nodeNames, newline,
+			"    MESSAGE: ", message, newline,
+		)
+		if err != nil {
+			return fmt.Errorf("failed to write out node status: %v", err)
+		}
+	default:
```
glibsm, line 341: i think default is better used for an "unknown type error" in case another enum is added

pkg/printer/printer.go
```diff
@@ -265,6 +267,99 @@ func (p *Printer) WriteProtoFlow(f v1.Flow) error {
...
+
+	switch p.opts.output {
+	case JSONOutput:
+		return json.NewEncoder(p.opts.werr).Encode(r)
+	case DictOutput:
+		// this is a bit crude, but in case stdout and stderr are interleaved,
+		// we want to make sure the separators are still printed to clearly
+		// separate flows from node events.
+		if p.line != 0 {
+			_, err := fmt.Fprintln(p.opts.w, "------------")
```
glibsm, line 320: share this const maybe? it's technically the length of the longest key, so may differ based on the message

cmd/observe/observe.go
```diff
@@ -408,12 +414,13 @@ func getFlows(client observer.ObserverClient, req *observer.GetFlowsRequest) err
 			return err
 		}
 
-		flow := getFlowResponse.GetFlow()
-		if flow == nil {
-			continue
+		if flow := getFlowResponse.GetFlow(); flow != nil {
```
michi-covalent, line 417: nit: might be cleaner to use type switch.

example: https://github.com/cilium/hubble/blob/cfcad093611faa9e26cbe1750998a8f5cab2b555/pkg/printer/printer.go#L167

michi-covalent:
@gandro i'm playing around with this branch, i see:
```
{
  "node_status": {
    "state_change": "NODE_CONNECTED",
    "node_names": [
      "gke-michikube0-default-pool-f424dd13-sfgw",
      "gke-michikube0-default-pool-f424dd13-xj5f",
      "gke-michikube0-default-pool-f424dd13-3k5t",
      "gke-michikube0-default-pool-f424dd13-wtvb"
    ]
  },
  "node_name": "hubble-relay-6954dddcdd-wsv7p",
  "time": "2020-05-27T23:48:24.967871213Z"
}
```
looks like `hubble-relay-6954dddcdd-wsv7p` is the name of the relay pod, not the node name. is this expected?

pkg/printer/printer.go
gandro: I enumerated the cases explicitly to be consistent with the Flow printer. So it will silently ignore node status for new enum variants. I would find it confusing one threw an error, but the other one didn't.

gandro:
Addressed the nits!

> looks like hubble-relay-6954dddcdd-wsv7p is the name of the relay pod, not the node name. is this expected?

It's somewhat expected in the way the node name is currently obtained, but I have not considered that K8s assigns the pod name as the host name. I guess if we wanted to change that we would have to modify the deployment of the hubble-relay pod. However, I think it's actually not a bad thing to have the pod name of the hubble-relay instance there as the source, since it is the one actually creating the status message on behalf of the nodes. The naming could be better, but changing the name of the `node_name` field in the `GetFlowsResponse` would be a breaking change. In any case, this is not something we can address on the client side in this PR ;)

michi-covalent:
ok guys i'm merging this

PR #268, opened 2020-05-28
make: optimize binary size by omitting symbol table and debug info

Rolinh:
Use the `-s -w` linker flags when compiling hubble in order to reduce the binary size.

The `-w` linker flag omits DWARF symbol table whereas the `-s` flag omits the symbol table and debug information.

This reduces the binary size from 18M to 13M.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ TEST_TIMEOUT ?= 5s
 all: hubble
 
 hubble:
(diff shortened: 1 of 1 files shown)


glibsm:
seems like we should probably have a `hubble-dev` tag or something for us specifically, that had pprof, debugging symbols, etc.

PR #269, opened 2020-05-28
cmd: Make pprof optional

gandro:
Since the hubble binary is only used as a client, there is not much
value in providing pprof support by default. This commit hides the
pprof logic behind a go build tag.

This reduces the binary size by ~150KB.


glibsm:
it starts.

PR #270, opened 2020-05-28
Ratelimit cilium identity fetch

### LINKED ISSUE #265: Rate-limits / Circuit-breakers to limit Cilium API calls made by Hubble
This issue is specific to Hubble with Cilium 1.7 and earlier, and is not relevant for embedded Hubble in Cilium 1.8 . 

Hubble fetches metadata (e.g., identity labels) for src + dst of each flow via the cilium APIs. This means the number of API calls it make to cilium can depend on the set of flows that appear on a node, and the identities associated with those flows.   Hubble caches this flow data, but scenarios like a large number of new or unknown identities can result in a potentially large number of Cilium API requests.    

A particularly bad scenario is where a flow may contain an IP for which cilium has no identity, which means hubble will not cache an identity, and that each flow with that IP may result in another API call.  

This issue is to explore whether there are simple improvements to the non-embedded version of hubble that can: 
(1) cap the overall number of requests that hubble makes to Cilium within a time window (or perhaps the max outstanding requests).  
(2) limit the impact of certain pathological cases, such as repeated lookups for an IP with no associated identity.  

Note, changes of both types may be ideal, as (1) will likely mean that we have to "fail hard" and not associate identity with flows when some cap is reached.  Thus, having something like (2) which can reduce the volume of potentially "high rate, low value" queries will help "protect" more valid lookups.

glibsm:
Fixes #265


pkg/parser/threefour/identity_backoff.go
```diff
@@ -0,0 +1,113 @@
...
+	"time"
+
+	"github.com/cilium/cilium/pkg/backoff"
+	hubtime "github.com/cilium/hubble/pkg/time"
+)
+
+const (
+	defaultBurst      = 25
+	defaultMinBackoff = 100 * time.Millisecond
+	defaultMaxBackoff = 5 * time.Second
```
glibsm, line 28: not super confident about these default settings, open to suggestions for burst/min/max

michi-covalent: max backoff can be larger i think, maybe 1 minute?

the most common scenario for missing identity i believe is when there are "stale" flows that contain deleted identities. in this case the identity will never be resolved, and we are basically waiting for stale flows to go away. i'm even fine without the default burst and immediately starting to back off :)

michi-covalent:
@glibsm btw i guess we will be creating v0.6 branch to go with cilium v1.8?

PR #271, opened 2020-05-28
observe: Remove --port-translation

michi-covalent:
This PR removes the --port-translation option, which uses gopacket to
look up port names. Removing the gopacket dependency reduces the binary
size by ~12%:

- Before: 13,508,608 bytes
- After : 11,915,264 bytes

Note that this change is not backward-compatible since it simply removes
the --port-translation option without any deprecation warning. I've
considered other less drastic approaches, but removing the option seems
like an acceptable solution here since this option doesn't seem to be
widely used.

Ref: https://godoc.org/github.com/google/gopacket/layers#TCPPort.String
Ref: https://godoc.org/github.com/google/gopacket/layers#UDPPort.String


Rolinh:
A less drastic approach that could still allow us to get rid of the `gopacket` dep is to use [a similar approach](https://github.com/google/gopacket/blob/d26664f9dcd3d61e328bc0bd3ad94eac7ce49ecb/layers/gen.go) to populate maps  with known ports name from IANA.

tgraf:
I'm in favor of this. I'm not sure port translation adds any meaningful value if it is not based on the port names in the pod/service spec.

glibsm:
> I'm in favor of this. I'm not sure port translation adds any meaningful value if it is not based on the port names in the pod/service spec.

that's an interesting idea

PR #289, opened 2020-06-04
Prepare v0.6.1 release

michi-covalent:


Makefile
```diff
@@ -4,7 +4,7 @@ BINDIR ?= /usr/local/bin
 IMAGE_REPOSITORY ?= quay.io/cilium/hubble
 CONTAINER_ENGINE ?= docker
 TARGET=hubble
-VERSION=0.6.1-dev
+VERSION=0.6.1
```
michi-covalent, line 7: am i supposed to set this to 0.6.2-dev at some point?

glibsm: https://github.com/cilium/hubble/blob/master/RELEASE.md#modify-the-version-constant-on-the-master-branch-if-needed

you set it on master branch post-release

michi-covalent: how about when i release a subsequent patch release (e.g. 0.6.1)? before i started this process, the Makefile had:
```
VERSION=0.6.1-dev
```
when 0.6.1 is ready, we do:
```
VERSION=0.6.1
```
but then after the release is done, shouldn't we set this to:
```
VERSION=0.6.2-dev
```
in `v0.6` branch? otherwise the hubble version will show up as `0.6.1` even after we add more commits to the branch.

btw somewhat related, GitBranch and GitHash aren't getting set when i run `make` from either master or v0.6 branch.
```
% make
go build  -ldflags "-w -s -X 'github.com/cilium/hubble/pkg.GitBranch=' -X 'github.com/cilium/hubble/pkg.GitHash=' -X 'github.com/cilium/hubble/pkg.Version=0.6.1-dev'" -o hubble
```
maybe it's a linux vs mac thing.

Rolinh: > maybe it's a linux vs mac thing.

I think it's rather that mac ships an antediluvian version of GNU make (ie: the last GPLv2 version released all the way back in 2006 which is 3.81 iirc) and it does not understand the `!=` operator that was added in GNU Make 4 and which executes and assigns the result to the variable. Let me cook a patch then.

Rolinh:
Let's block this PR until #290 is merged and backported to the `v0.6` branch.

PR #290, opened 2020-06-05
make: fix git hash variable assignments for old make versions

Rolinh:
Antediluvian versions of GNU make, such as the one shipped by default on
MacOS (GNU Make 3.81, the last GPLv2 version released back in 2006) do
not understand the `!=` operator which executes and assigns the results
to a variable. This results in the GIT_BRANCH and GIT_HASH variables not
being assigned on MacOS. This commit addresses this issue by using the
`$(shell ...)` construct instead.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,8 +5,8 @@ IMAGE_REPOSITORY ?= quay.io/cilium/hubble
 CONTAINER_ENGINE ?= docker
 TARGET=hubble
 VERSION=0.7.0-dev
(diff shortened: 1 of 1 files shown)


Rolinh:
> This looks suspiciously like something I've already fixed before... maybe in another repo, or in another life

We talked about it offline ~1 month ago, that's probably why. I was also sure that one of us had already commited a fix but... nope.

PR #291, opened 2020-06-05
make: fix git hash variable assignments for old make versions

michi-covalent:
Antediluvian versions of GNU make, such as the one shipped by default on
MacOS (GNU Make 3.81, the last GPLv2 version released back in 2006) do
not understand the `!=` operator which executes and assigns the results
to a variable. This results in the GIT_BRANCH and GIT_HASH variables not
being assigned on MacOS. This commit addresses this issue by using the
`$(shell ...)` construct instead.

Signed-off-by: Robin Hahling <robin.hahling@gw-computing.net>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,8 +5,8 @@ IMAGE_REPOSITORY ?= quay.io/cilium/hubble
 CONTAINER_ENGINE ?= docker
 TARGET=hubble
 VERSION=0.6.1-dev
-GIT_BRANCH != which git >/dev/null 2>&1 && git rev-parse --abbrev-ref HEAD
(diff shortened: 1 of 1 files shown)


michi-covalent:
tested on mac:
```
% make
go build  -ldflags "-w -s -X 'github.com/cilium/hubble/pkg.GitBranch=pr/michi/backportgogo' -X 'github.com/cilium/hubble/pkg.GitHash=848ff00' -X 'github.com/cilium/hubble/pkg.Version=0.6.1-dev'" -o hubble
% ./hubble version
hubble v0.6.1-dev@pr/michi/backportgogo-848ff00 compiled with go1.14 on darwin/amd64
```

PR #295, opened 2020-06-10
Build release artifacts inside a container

michi-covalent:
- Use make clean to delete the release directory.
- Run `make release` from inside a docker container to ensure that the
  artifacts get generated in a reproducible environment.
- Generate release artifacts as a part of the build.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -28,3 +28,11 @@ jobs:
         make check
     - name: Run unit tests
       run: make test
+    - name: Generate artifacts
+      run: ./contrib/scripts/release.sh
+    - name: Upload artifacts
+      uses: actions/upload-artifact@v2
+      if: success()
+      with:
+        name: release
+        path: release
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -111,9 +111,9 @@ https://github.com/cilium/hubble/releases/new
 
 ## Finally, upload release tarballs to the GitHub release
 
-Generate the release tarballs using the `release` make target:
+Generate the release tarballs using `contrib/scripts/release.sh` script:
 
-    make release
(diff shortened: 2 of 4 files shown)


.github/workflows/unittests.yml
```diff
@@ -28,3 +28,11 @@ jobs:
         make check
     - name: Run unit tests
       run: make test
+    - name: Generate artifacts
+      run: ./contrib/scripts/release.sh
+    - name: Upload artifacts
```
Rolinh, line 33: I understand the benefit of running the `release.sh` script but I don't understand the benefit of uploading the artifacts.
I would understand that we automatically build and upload artifacts on the release page when a release is tagged but this is not what this is doing, is it?

gandro: This will allow us to download the binaries from each commit for testing. Which I think is a nice feature, as we might want to use these artifacts for the official releases as well.

michi-covalent:
backporting to v0.6 https://github.com/cilium/hubble/pull/296

RELEASE.md
```diff
@@ -111,9 +111,9 @@ https://github.com/cilium/hubble/releases/new
 
 ## Finally, upload release tarballs to the GitHub release
 
-Generate the release tarballs using the `release` make target:
+Generate the release tarballs using `contrib/scripts/release.sh` script:
 
-    make release
+    ./contrib/scripts/release.sh
```
glibsm, line 116: imo `make release` should run the docker build. current release can become some other target

i find it easier to run `make release` than to remember what script and where should be run (especially if we refactor it again)

michi-covalent: fair point, will address in a follow up pr.

michi-covalent: https://github.com/cilium/hubble/pull/297

PR #296, opened 2020-06-11
Build release artifacts inside a container

michi-covalent:
- Use make clean to delete the release directory.
- Run `make release` from inside a docker container to ensure that the
  artifacts get generated in a reproducible environment.
- Generate release artifacts as a part of the build.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -28,3 +28,11 @@ jobs:
         make check
     - name: Run unit tests
       run: make test
+    - name: Generate artifacts
+      run: make release
+    - name: Upload artifacts
+      uses: actions/upload-artifact@v2
+      if: success()
+      with:
+        name: release
+        path: release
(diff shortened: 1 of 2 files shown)


.github/workflows/unittests.yml
```diff
@@ -28,3 +28,11 @@ jobs:
         make check
     - name: Run unit tests
       run: make test
+    - name: Generate artifacts
+      run: ./contrib/scripts/release.sh
+    - name: Upload artifacts
+      uses: actions/upload-artifact@v2
+      if: success()
```
glibsm, line 35: does it have to be `if: ${{ success() }}`?

michi-covalent: no, amazing syntactic sugar:

> When you use expressions in an if conditional, you may omit the expression syntax (${{ }}) because GitHub automatically evaluates the if conditional as an expression.

https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#about-contexts-and-expressions

glibsm: I also followed their documentation and found the non-omition syntax in the example https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions

glibsm:
I imagine we're going to wait on merging this to also backport whatever is the result of https://github.com/cilium/hubble/pull/297

michi-covalent:
i'll squash these 2 commits

Issue #298, opened 2020-06-12
proposal: managing recommended hubble version in cilium doc


michi-covalent:
# problem

right now the documentation points to the latest release https://github.com/cilium/hubble/releases/latest. this is problematic because the link simply points to the most recent release, and we cannot control which version to recommend in the cilium documentation. for example, we don't want to point to a new patch release of an older minor version.

also, we don't want to be tweaking the cilium documentation constantly just to update the recommended hubble version.

# proposal

introduce `stable.txt` that contains the recommended version, and change the download instruction to:
```
curl -LO https://github.com/cilium/hubble/releases/download/`curl -s https://raw.githubusercontent.com/cilium/hubble/master/stable.txt`/hubble-linux-amd64.tar.gz
```

# details

at the very least, the hubble version specified in `stable.txt` should be compatible with the stable version of cilium. assuming that we maintain backward compatibility, we should be able to reuse the same installation instructions in the previous documentation versions. 

on (hopefully) rare occasion we break backward compatibility, the installation instructions in older documentation versions need to be updated to point to a specific version of hubble.

PR #299, opened 2020-06-12
Add stable.txt

michi-covalent:
Point stable.txt to v0.6.1 and update RELEASE.md. For now there is a
single stable.txt since there is only one Cilium version to support
and we haven't broken backward compatibility.

Ref #298


RELEASE.md
```diff
@@ -118,3 +118,16 @@ Generate the release tarballs using `contrib/scripts/release.sh` script:
 This will generate tarballs and associated checksum files in the `release`
 directory. Make sure to upload these tarball and checksum to the GitHub release
 page.
+
+## (OPTIONAL) Update `stable.txt` in the master branch
```
glibsm, line 122: just how optional?

michi-covalent: it's somewhat fuzzy right now. if you are releasing the latest version (say v0.6.2), you most likely want to update `stable.txt` so that people start picking up bug fixes and what not.

if you are releasing a patch version of older release (v0.5.2 for example), you don't want update `stable.txt` because it's already pointing to v0.6.1.

glibsm: Btw, this is good info that provides a lot more context than `optional`

michi-covalent: agreed. will update the section

michi-covalent: updated RELEASE.md

PR #301, opened 2020-06-15
v0.5: Pin Hubble and Cilium branches

gandro:
This PR ensures that we do not refer to the Hubble `master` branch in our Hubble v0.5 docs, as in Hubble 0.6 and latest, those files (i.e. Helm charts and documentation) have moved to the Cilium repository and will soon be removed from the Hubble master branch.

In addition, the PR also replaces a few occurrences of the `v1.7.0` Cilium git tag with the `v1.7` Cilium git branch, which will pick up the latest Cilium 1.7 patch release.

### DIFF
--- a/Documentation/installation.md
+++ b/Documentation/installation.md
@@ -7,9 +7,9 @@
 
 ## Install Cilium
 
-Install Cilium using the [Install instructions]. To deploy Cilium 1.7.0 using quick-install.yaml:
+Install Cilium using the [Install instructions]. To deploy Cilium 1.7 using quick-install.yaml:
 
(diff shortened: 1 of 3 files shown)


Documentation/installation.md
```diff
@@ -2,14 +2,14 @@
 
 ## Requirements
 
- * [Cilium] Recommended: >= 1.7.0, Minimal: >= 1.6.3
+ * [Cilium] >= 1.7.0
```
glibsm, line 5: <1.8?

michi-covalent: yeah maybe better to limit the range to `1.7.*`

gandro: Good point! I just back ported the commit we applied to master, but explicitly indicating we don't support Hubble 0.5 on Cilium 1.8 is a good call.

PR #302, opened 2020-06-15
printer: Add jsonpb output

michi-covalent:
This output option serializes the entire GetFlowResponse using proto3's
JSON mapping so that the JSON format is consistent across all the event
types. The output looks like this:

```
% hubble observe --debug -o jsonpb 2>&1 | jq
{
  "node_status": {
    ...
  },
  "node_name": "hubble-relay-565f76c75f-974d9",
  "time": "2020-06-15T20:56:44.689246261Z"
}
{
  "flow": {
    ...
  },
  "node_name": "hubble-relay-565f76c75f-974d9",
  "time": "2020-06-15T20:56:44.689246262Z"
}
```

Here is the updated description of `--output` option:
```
-o, --output string      Specify the output format, one of:
                          compact:  Compact output
                          dict:     Each flow is shown as KEY:VALUE pair
                          json:     JSON encoding
                          jsonpb:   Output each GetFlowResponse according to proto3's JSON mapping
                          table:    Tab-aligned columns
```

### DIFF
--- a/cmd/observe/observe.go
+++ b/cmd/observe/observe.go
@@ -43,6 +43,7 @@ var (
 	sinceVar, untilVar string
 
 	jsonOutput          bool
+	jsonPBOutput        bool
 	compactOutput       bool
 	dictOutput          bool
 	output              string
@@ -219,8 +220,12 @@ programs attached to endpoints and devices. This includes:
 	)
 	observerCmd.Flags().StringVarP(
 		&output, "output", "o", "",
-		"Specify the output format, one of:\n -compact:\tcompact output\n -dict:\teach flow is shown as KEY:VALUE pair\n -json:\tJSON encoding\n -table:\ttab-aligned columns",
-	)
+		`Specify the output format, one of:
+ compact:  Compact output
+ dict:     Each flow is shown as KEY:VALUE pair
+ json:     JSON encoding
+ jsonpb:   Output each GetFlowResponse according to proto3's JSON mapping
+ table:    Tab-aligned columns`)
 	observerCmd.Flags().BoolVarP(
 		&ignoreStderr, "silent-errors", "s", false, "Silently ignores errors and warnings")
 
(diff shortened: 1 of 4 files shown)


cmd/observe/observe.go
```diff
@@ -219,8 +220,12 @@ programs attached to endpoints and devices. This includes:
 	)
 	observerCmd.Flags().StringVarP(
 		&output, "output", "o", "",
-		"Specify the output format, one of:\n -compact:\tcompact output\n -dict:\teach flow is shown as KEY:VALUE pair\n -json:\tJSON encoding\n -table:\ttab-aligned columns",
-	)
+		`Specify the output format, one of:
+ compact:  Compact output
+ dict:     Each flow is shown as KEY:VALUE pair
+ json:     Deprecated. Use '--output raw-json' instead.
```
Rolinh, line 226: Why is `json` being deprecated?

michi-covalent: i can go either way, but i wanted to deprecate this for one specific reason. right now the Flow type has node_name and time fields. these fields are redundant because they are also included at the top level.

if we can deprecate this current json output format, it also makes it easier to deprecate node_name/time fields inside the Flow type.

another aspect is that i simply wanted to avoid having 2 similar json output formats.

for now i'll remove the deprecation warning. let's have a separate discussion.

glibsm:
Can you paste a diff between the two outputs?

michi-covalent:
the only different is that `-o json` only prints what's inside the `flow` field.

[json.txt](https://github.com/cilium/hubble/files/4787891/json.txt)
[jsonpb.txt](https://github.com/cilium/hubble/files/4787892/jsonpb.txt)

glibsm:
Got it, so:

```
{
  "flow": {
    .... jsonpb output
  },
  "node_name": "minikube",
  "time": "2020-06-16T16:39:06.302589475Z"
}
```

michi-covalent:
it's the other way around. `-o jsonpb` prints:
```
{
  "flow": {
    .... json output
  },
  "node_name": "minikube",
  "time": "2020-06-16T16:39:06.302589475Z"
}
```

glibsm:
Disappointed this wasn't called `-o pbj` :D

PR #311, opened 2020-06-29
ghactions: Do not upload artifacts for each commit

gandro:
Uploading the release artifacts for each commit seems to eat away at the
GitHub shared storage and bandwith quota of the Cilium GitHub organization.
Since we are not actively using these artifacts, this commit disables the
upload to avoid wasting storage space.


gandro:
Closing this again. Seems like we misread the data. This does not affect our LFS storage/bandwidth quota, which is currently running low.

PR #317, opened 2020-07-10
.gitattributes: hide go.sum and vendor/modules.txt in pull requests

Rolinh:
This marks the go.sum and vendor/modules.txt file sas generated so that they are collapsed by default in Github pull requests.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+go.sum linguist-generated
+vendor/modules.txt linguist-generated


.gitattributes
```diff
@@ -0,0 +1 @@
+go.sum linguist-generated
```
glibsm, line 1: ` vendor/modules.txt` would be nice to hide as well

PR #318, opened 2020-07-13
observe: Document default flow count output

joestringer:
I couldn't tell from the help text what the intended default was here,
but there seems to be some special logic to try to ensure that the
number of flows is only restricted when no other selector parameters are
specified. Manually add the default string at the end so that we don't
have to restrict the number of flows to 20 by default when specifying
other selector criteria, but it's still apparent what the default
behaviour is.

### DIFF
--- a/cmd/observe/observe.go
+++ b/cmd/observe/observe.go
@@ -107,7 +107,7 @@ programs attached to endpoints and devices. This includes:
 		"type", "t", ofilter, []string{},
 		fmt.Sprintf("Filter by event types TYPE[:SUBTYPE] (%v)", eventTypes())))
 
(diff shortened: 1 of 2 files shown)


glibsm:
`last` and `since|until` are mutually exclusive. Even if you provide other filters such as `--from-ip=12.31.231.23`, you will still only get back 20.

That said I think your approach is ok to use here instead of the default helpers for cobra since we'd like to know if the user actually set it to 20, or that's just the default (makes error handling easier without propagating the default value everywhere).

joestringer:
> `last` and `since`|`until` are mutually exclusive. Even if you provide other filters such as `--from-ip=12.31.231.23`, you will still only get back 20.

Ah cool, I wasn't sure but I figured I'd go for the minimal change here since I'm sure it was written this way for a reason.

PR #319, opened 2020-07-15
update Go version to v1.14.5

Rolinh:


Rolinh:
> does version in go.mod matter?

The [go directive](https://golang.org/ref/mod#tmp_12) in the `go.mod` file indicates the minimum Go version for the project (it defaults to the current version if not specified). So I don't think we want to bump it up unless we have to.

PR #321, opened 2020-07-21
make: fix release build directory ownership

kAworu:
Before this patch, `make release` would generate the release/ directory (and files underneath) owned by root:root. This is caused by the fact that the build happen as root inside the Docker container. As a result, a following `make clean` from outside the container as an unprivileged user would fail.

This patch fix the issue by chowning the release directory and files to the user uid and gid whom invoked make originally (i.e. from outside the container).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,8 @@ VERSION=0.7.0-dev
 GIT_BRANCH = $(shell which git >/dev/null 2>&1 && git rev-parse --abbrev-ref HEAD)
 GIT_HASH = $(shell which git >/dev/null 2>&1 && git rev-parse --short HEAD)
 GO_TAGS ?=
+RELEASE_UID ?= $(shell id -u)
+RELEASE_GID ?= $(shell id -g)
 
 TEST_TIMEOUT ?= 5s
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -30,9 +32,12 @@ local-release: clean
 			test -d release/$$OS/$$ARCH|| mkdir -p release/$$OS/$$ARCH; \
 			env GOOS=$$OS GOARCH=$$ARCH $(GO) build $(if $(GO_TAGS),-tags $(GO_TAGS)) -ldflags "-w -s -X 'github.com/cilium/hubble/pkg.Version=${VERSION}'" -o release/$$OS/$$ARCH/$(TARGET)$$EXT; \
 			tar -czf release/$(TARGET)-$$OS-$$ARCH.tar.gz -C release/$$OS/$$ARCH $(TARGET)$$EXT; \
-			cd release && sha256sum $(TARGET)-$$OS-$$ARCH.tar.gz > $(TARGET)-$$OS-$$ARCH.tar.gz.sha256sum && cd $(CURDIR); \
+			(cd release && sha256sum $(TARGET)-$$OS-$$ARCH.tar.gz > $(TARGET)-$$OS-$$ARCH.tar.gz.sha256sum); \
 		done; \
 		rm -r release/$$OS; \
+		if [ $$(id -u) -eq 0 -a -n "$$RELEASE_UID" -a -n "$$RELEASE_GID" ]; then \
+			chown -R "$$RELEASE_UID:$$RELEASE_GID" release; \
+		fi; \
```
Rolinh, line 40: What is the reason for doing it as part of the `for` loop? I think this could be done once after the loop instead.

kAworu: Good catch, my intention was to do it after the loop but I overlooked the outer OS loop.

Rolinh: Would you mind squashing the two commits?

kAworu: Will do

Makefile
```diff
@@ -30,10 +32,13 @@ local-release: clean
 			test -d release/$$OS/$$ARCH|| mkdir -p release/$$OS/$$ARCH; \
 			env GOOS=$$OS GOARCH=$$ARCH $(GO) build $(if $(GO_TAGS),-tags $(GO_TAGS)) -ldflags "-w -s -X 'github.com/cilium/hubble/pkg.Version=${VERSION}'" -o release/$$OS/$$ARCH/$(TARGET)$$EXT; \
 			tar -czf release/$(TARGET)-$$OS-$$ARCH.tar.gz -C release/$$OS/$$ARCH $(TARGET)$$EXT; \
-			cd release && sha256sum $(TARGET)-$$OS-$$ARCH.tar.gz > $(TARGET)-$$OS-$$ARCH.tar.gz.sha256sum && cd $(CURDIR); \
+			(cd release && sha256sum $(TARGET)-$$OS-$$ARCH.tar.gz > $(TARGET)-$$OS-$$ARCH.tar.gz.sha256sum); \
```
Rolinh, line 35: :exploding_head:

PR #328, opened 2020-07-27
cmd/status: print node availability information when available

Rolinh:
Example output when all nodes are connected:

    $ hubble status
    Max Flows: 81920
    Current Flows: 24700 (30.15%)
    Connected Nodes: 20/20

Example output with unavailable nodes:

    $ hubble status
    Max Flows: 24576
    Current Flows: 24576 (100.00%)
    Connected Nodes: 6/20
    Unavailable Nodes: 14
      - kind-worker10
      - kind-worker11
      - kind-worker12
      - kind-worker14
      - kind-worker15
      - kind-worker16
      - kind-worker18
      - kind-worker7
      - kind-worker8
      - kind-worker9
      - and 4 more...

Note: for this to work as above, this server-side fix for Hubble Relay is required: https://github.com/cilium/cilium/pull/12685


Rolinh:
I wonder if it makes sense to print info about unavailable nodes when they all are available. Thoughts?

glibsm:
> I wonder if it makes sense to print info about unavailable nodes when they are all available. Thoughts?

I think we can omit `Unavailable nodes: 0` when they are all healthy, yeah.

cmd/status/status.go
```diff
@@ -70,6 +72,29 @@ func runStatus() error {
...
+	if numConnected != nil {
+		total := ""
+		if numUnavailable != nil {
+			total = fmt.Sprintf("/%d", numUnavailable.Value+numConnected.Value)
+		}
+		fmt.Printf("Connected nodes: %d%s\n", numConnected.Value, total)
+	}
+	if numUnavailable != nil {
+		if unavailable := ss.GetUnavailableNodes(); unavailable != nil && numUnavailable.Value > 0 {
+			sort.Strings(unavailable) // it's nicer when displaying unavailable nodes list
```
glibsm, line 86: why would the server not sort this for all clients?

cmd/status/status.go
```diff
@@ -70,6 +72,29 @@ func runStatus() error {
...
+			if numUnavailable.Value > uint32(len(unavailable)) {
+				unavailable = append(unavailable, fmt.Sprintf("and %d more...", numUnavailable.Value-uint32(len(unavailable))))
+			}
+			fmt.Printf("Unavailable nodes: %d\n  - %s\n",
+				numUnavailable.Value,
+				strings.Join(unavailable, "\n  - "),
```
glibsm, line 92: this is clever in one statement

cmd/status/status.go
```diff
@@ -70,6 +72,29 @@ func runStatus() error {
...
+		}
+		fmt.Printf("Connected nodes: %d%s\n", numConnected.Value, total)
+	}
+	if numUnavailable != nil {
+		if unavailable := ss.GetUnavailableNodes(); unavailable != nil && numUnavailable.Value > 0 {
+			sort.Strings(unavailable) // it's nicer when displaying unavailable nodes list
+			if numUnavailable.Value > uint32(len(unavailable)) {
+				unavailable = append(unavailable, fmt.Sprintf("and %d more...", numUnavailable.Value-uint32(len(unavailable))))
+			}
+			fmt.Printf("Unavailable nodes: %d\n  - %s\n",
```
gandro, line 90: nit: We have "Max/Current Flows" (capital case). Should we have "Unavailable Nodes" here as well?

Rolinh: Yup, fixed.

cmd/status/status.go
Rolinh: Good point, this can certainly be done.

Rolinh:
I reverted the cilium update to the commit just before https://github.com/cilium/cilium/commit/1d1427b4bb89208dc134d0b4afcce9c323042606 as this change breaks the tests. It looks like it's going to be something tricky to fix :confused:

PR #330, opened 2020-07-28
cmd/status: Add flows per second to `hubble status`

### LINKED ISSUE #329: Adds flows/s to hubble status
flows/s is included in `cilium status`, but not `hubble status`

```
❯ kubectl exec -n kube-system -t ds/cilium -- cilium status | grep Hubble
Hubble:                 Ok              Current/Max Flows: 4096/4096 (100.00%), Flows/s: 21.16   Metrics: Ok
❮ kubectl exec -n kube-system -t ds/cilium -- hubble status
Healthcheck (via unix:///var/run/cilium/hubble.sock): Ok
Max Flows: 4096
Current Flows: 4096 (100.00%)
```

gandro:
This information is also provided by `cilium status` in Cilium 1.8 and newer.

```console
$ hubble status
Healthcheck (via unix:///var/run/cilium/hubble.sock): Ok
Max Flows: 4096
Current Flows: 4096 (100.00%) 
Flows/s: 9.20
```

Formatting is up for discussion. I opted to put it on a new line, but there are other variants, such as e.g. `Current Flows: 4096 (100%, 9.20 flows/s)`

Fixes #329

### DIFF
--- a/cmd/status/status.go
+++ b/cmd/status/status.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/cilium/cilium/api/v1/observer"
 	v1 "github.com/cilium/cilium/pkg/hubble/api/v1"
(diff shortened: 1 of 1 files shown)


cmd/status/status.go
```diff
@@ -72,6 +73,10 @@ func runStatus() error {
 		ss.NumFlows,
 		(float64(ss.NumFlows)/float64(ss.MaxFlows))*100,
 	)
+	if ss.UptimeNs > 0 {
+		fmt.Printf("Flows/s: %.2f\n", float64(ss.SeenFlows)/time.Duration(ss.UptimeNs).Seconds())
```
glibsm, line 77: i'm assuming unlikely, but still.. divide by zero guard?

gandro: I guess I went for the same approach as we did with the unavailable nodes: If it's not part of the status message (i.e. Hubble v0.5 server) we omit it. But I can add the `N/A`.

As for the guard: Good catch. I guess it won't cause a crash (non-signalling float math), but `Flows/s: Inf` will not be a good result either :D

gandro: Addressed.

gandro: Oh, actually, 1 ns is still too large ever cause an float underflow. So the previous check could never fail as well, but let's guard on the `Seconds()` anyways.

PR #334, opened 2020-07-31
Allow enabling the hubble-grpc service without Hubble UI enabled

kAworu:
_**NOTE**: This PR targets the v0.5 branch._

Allow users to enable the hubble-grpc Service (through `.Values.grpc.enabled`) without requiring Hubble UI to be enabled. Note that when Hubble UI is enabled, the hubble-grpc Service is enabled unconditionally since UI depends on it.

### DIFF
--- a/install/kubernetes/hubble/Chart.yaml
+++ b/install/kubernetes/hubble/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: hubble
-version: 0.0.3
-appVersion: 0.0.1
+version: 0.5.2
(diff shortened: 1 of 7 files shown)


install/kubernetes/hubble/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: hubble
-version: 0.0.3
-appVersion: 0.0.1
+version: 0.5.2
+appVersion: 0.5.2
```
Rolinh, line 4: Nice catch!

PR #335, opened 2020-08-05
Update Cilium dep and fix unit tests that subsequently broke

Rolinh:
Bump Cilium dep to latest and use `go-cmp` to rewrite assertion that started to fail after a cilium dep upgrade due to a `testify` bug.

This pulls in `github.com/google/go-cmp` as an explicit dependency.

Ref: https://github.com/cilium/cilium/issues/12772


cmd/observe/observe_filter_test.go
```diff
@@ -86,20 +88,31 @@ func TestFilterDispatch(t *testing.T) {
...
+			},
 		},
-	}, f.whitelist.flowFilters(), "whitelist filter is incorrect")
-
-	assert.Equal(t, []*pb.FlowFilter{
-		{
-			DestinationIp: []string{"5.5.5.5"},
+		f.whitelist.flowFilters(),
+		cmpopts.IgnoreUnexported(pb.FlowFilter{}),
+		cmpopts.IgnoreUnexported(pb.EventTypeFilter{}),
```
glibsm, line 101: is this just future-proofing, or the test doesn't pass without these?

Rolinh: When omitted, one gets a panic with the following message:
```
--- FAIL: TestFilterDispatch (0.00s)
panic: cannot handle unexported field at {[]*flow.FlowFilter}[0].EventType[0].state:
        "github.com/cilium/cilium/api/v1/flow".EventTypeFilter
consider using a custom Comparer; if you control the implementation of type, you can also consider using an Exporter, AllowUnexported, or cmpopts.IgnoreUnexported [recovered]
        panic: cannot handle unexported field at {[]*flow.FlowFilter}[0].EventType[0].state:
        "github.com/cilium/cilium/api/v1/flow".EventTypeFilter
consider using a custom Comparer; if you control the implementation of type, you can also consider using an Exporter, AllowUnexported, or cmpopts.IgnoreUnexported
```

PR #343, opened 2020-08-12
Update Go to v1.15, drop support for darwin/386, add support for linux/[arm,arm64]

Rolinh:
See commits for details.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -17,7 +17,7 @@ jobs:
         path: go/src/github.com/cilium/hubble
     - uses: actions/setup-go@v1
       with:
-        go-version: '1.14.7'
+        go-version: '1.15'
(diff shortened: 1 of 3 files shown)


.github/workflows/unittests.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         path: go/src/github.com/cilium/hubble
     - uses: actions/setup-go@v1
       with:
-        go-version: '1.14.7'
+        go-version: '1.15'
```
glibsm, line 20: does this imply `1.15.0`?

Rolinh: Good point. I assumed that this was following the Go version scheme where `1.15` means `1.15.0`. To make sure, I changed it to `1.15.0` and go a failure for the Github Action so then reverted back to `1.15`.

PR #344, opened 2020-08-12
Add staticcheck to `make check`

tklauser:
Add `staticcheck` to the `make check` target which will also run it as part of GitHub actions.

The first two commits fix remaining issues reported by `staticcheck`, the third commit enables the check. Currently, the following checks are disabled:

* SA1019 deprecation warnings: https://staticcheck.io/docs/checks#SA1019
* ST1000 missing package comment: https://staticcheck.io/docs/checks#ST1000

I guess the second one should eventually be fixed by adding package level godoc comments to all sub-packages.

### DIFF
--- a/pkg/printer/printer.go
+++ b/pkg/printer/printer.go
@@ -92,7 +92,7 @@ func (p *Printer) Close() error {
 
 // WriteErr returns the given msg into the err writer defined in the printer.
 func (p *Printer) WriteErr(msg string) error {
-	_, err := fmt.Fprint(p.opts.werr, fmt.Sprintf("%s\n", msg))
+	_, err := fmt.Fprintf(p.opts.werr, "%s\n", msg)
 	return err
 }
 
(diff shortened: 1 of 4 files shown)


pkg/printer/printer.go
```diff
@@ -92,7 +92,7 @@ func (p *Printer) Close() error {
 
 // WriteErr returns the given msg into the err writer defined in the printer.
 func (p *Printer) WriteErr(msg string) error {
-	_, err := fmt.Fprint(p.opts.werr, fmt.Sprintf("%s\n", msg))
+	_, err := fmt.Fprintf(p.opts.werr, "%s\n", msg)
```
glibsm, line 95: `fmt.Fprintln(p.opts.werr, msg`?

tklauser: Yeah, even better. Was following `staticcheck`'s advise a bit too closely. Will send a follow-up PR.

PR #355, opened 2020-08-25
Dockerfile: Remove ENTRYPOINT

michi-covalent:
Specifying ENTRYPOINT doesn't make sense for this image anymore since
it no longer runs Hubble server, and it's primarily used to access the
CLI. Remove ENTRYPOINT so that you can bash into the container without
providing additional arguments to override it.


Rolinh (approved):

gandro (approved):
Works like a charm!

PR #356, opened 2020-08-28
make: use command -v instead of which

kAworu:
Before this patch, `which(1)` failure would be printed to stderr even when
a target unrelated the the failure would be run:

```
% make hubble
which: no staticcheck in (/home/alex/go/bin:/sbin:/bin:/usr/sbin:/usr/bin:/usr/games:/usr/local/sbin:/usr/local/bin)
go build  -ldflags "-w -s -X 'github.com/cilium/hubble/pkg.GitBranch=pr/kaworu/makefile-command-instead-of-which' -X 'github.com/cilium/hubble/pkg.GitHash=c9707b0' -X 'github.com/cilium/hubble/pkg.Version=0.7.0-dev'" -o hubble
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,8 +5,8 @@ IMAGE_REPOSITORY ?= quay.io/cilium/hubble
 CONTAINER_ENGINE ?= docker
 TARGET=hubble
 VERSION=0.7.0-dev
-GIT_BRANCH = $(shell which git >/dev/null 2>&1 && git rev-parse --abbrev-ref HEAD)
-GIT_HASH = $(shell which git >/dev/null 2>&1 && git rev-parse --short HEAD)
+GIT_BRANCH = $(shell command -v git >/dev/null 2>&1 && git rev-parse --abbrev-ref HEAD)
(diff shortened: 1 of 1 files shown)


Rolinh (approved):

Rolinh:
I wonder though if we should not simply vendor `ineffasign`, `golint` and `staticcheck` (eg with blank imports).

kAworu:
@Rolinh then we would have the binaries in the `vendor` directory? How different CPU architectures would be handled (e.g. when we make from aarch64 for example)?

Rolinh:
> How different CPU architectures would be handled (e.g. when we make from aarch64 for example)?

That's a good point. Can't we just have the source code and use `go run ...` ?

glibsm:
> @Rolinh then we would have the binaries in the `vendor` directory? How different CPU architectures would be handled (e.g. when we make from aarch64 for example)?

Usually one would vendor just the source code and shim over with a `go run` like we do with https://github.com/cilium/cilium/blob/master/tools/protoc-gen-gogo for example

kAworu:
follow-up upon @glibsm suggestion in #357

PR #357, opened 2020-08-31
make: vendor in ineffassign, staticcheck, and golint

kAworu:
Vendor in tools used by make so that we don't rely on the host machine to have these installed in `$PATH`.

See also #356


Rolinh (changes requested):
Nice, lgtm! Please, update `.github/workflows/unittests.yml` as well.

glibsm:
Nice! As @Rolinh said we can drop some go gets in the unit test runs

Rolinh (approved):

Issue #86, opened 2020-02-09
Useless node name in CLI output


tgraf:
The node name in the output is pretty useless when operating the CLI in a single node context and only steals space.

```
ks exec -ti hubble-gt7m2 -- hubble observe -t drop --follow
Feb  9 13:28:21.965 [ip-192-168-12-253.us-west-2.compute.internal]: default/pod-to-a-l3-denied-cnp-65fd99db76-2knrs:51104 -> default/echo-a-7f74df756d-s4b9p:80(http) Policy denied (L3) DROPPED (TCP Flags: SYN)
Feb  9 13:28:22.798 [ip-192-168-12-253.us-west-2.compute.internal]: fe80::3042:1dff:fe0c:17df -> ff02::2 Unsupported L3 protocol DROPPED (ICMPv6 RouterSolicitation)
Feb  9 13:28:23.743 [ip-192-168-12-253.us-west-2.compute.internal]: default/pod-to-a-l3-denied-cnp-65fd99db76-2knrs:51112 -> default/echo-a-7f74df756d-s4b9p:80(http) Policy denied (L3) DROPPED (TCP Flags: SYN)
Feb  9 13:28:23.982 [ip-192-168-12-253.us-west-2.compute.internal]: default/pod-to-a-l3-denied-cnp-65fd99db76-2knrs:51104 -> default/echo-a-7f74df756d-s4b9p:80(http) Policy denied (L3) DROPPED (TCP Flags: SYN)
Feb  9 13:28:24.749 [ip-192-168-12-253.us-west-2.compute.internal]: default/pod-to-a-l3-denied-cnp-65fd99db76-2knrs:51112 -> default/echo-a-7f74df756d-s4b9p:80(http) Policy denied (L3) DROPPED (TCP Flags: SYN)
Feb  9 13:28:26.765 [ip-192-168-12-253.us-west-2.compute.internal]: default/pod-to-a-l3-denied-cnp-65fd99db76-2knrs:51112 -> default/echo-a-7f74df756d-s4b9p:80(http) Policy denied (L3) DROPPED (TCP Flags: SYN)
Feb  9 13:28:30.222 [ip-192-168-12-253.us-west-2.compute.internal]: fe80::3042:1dff:fe0c:17df -> ff02::2 Unsupported L3 protocol DROPPED (ICMPv6 RouterSolicitation)
```

genbit:
[...]

PR #358, opened 2020-08-31
printer: ommit node name from output

mdnix:
The node name gets ommited from the compact output by default since it takes up too much space. The node name can be displayed by adding the --node flag

Signed-off-by: Marco De Luca <marcodl404@gmail.com>

### DIFF
--- a/cmd/observe/observe.go
+++ b/cmd/observe/observe.go
@@ -49,6 +49,7 @@ var (
 	follow              bool
 	ignoreStderr        bool
 	enableIPTranslation bool
+	nodeName            bool
 
 	printer *hubprinter.Printer
 
@@ -242,6 +243,8 @@ programs attached to endpoints and devices. This includes:
 		"Translate IP addresses to logical names such as pod name, FQDN, ...",
 	)
 
+	observerCmd.Flags().BoolVarP(&nodeName, "print-node-name", "", false, "Print node name in output")
+
 	customObserverHelp(observerCmd)
 
 	return observerCmd
@@ -303,6 +306,9 @@ func handleArgs(ofilter *observeFilter) (err error) {
 	}
 	if debug {
 		opts = append(opts, hubprinter.WithDebug())
+	}
+	if nodeName {
+		opts = append(opts, hubprinter.WithNodeName())
 	}
 	printer = hubprinter.New(opts...)
 	return nil
--- a/pkg/printer/printer.go
+++ b/pkg/printer/printer.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net"
 	"os"
 	"path"
@@ -41,6 +42,18 @@ type Printer struct {
 	line        int
 	tw          *tabwriter.Writer
 	jsonEncoder *json.Encoder
+}
+
+type errWriter struct {
+	w   io.Writer
+	err error
+}
+
+func (ew *errWriter) write(a ...interface{}) {
+	if ew.err != nil {
+		return
+	}
+	_, ew.err = fmt.Fprint(ew.w, a...)
 }
 
 // New Printer.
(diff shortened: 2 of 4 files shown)


glibsm:
It's a bit of a harder sell to change the default behaviour. What about the opposite -- an option to omit the node name?

mdnix:
We could sure do that. Would actually make more sense since it wouldn't change anything in the default behaviour.
@Rolinh also suggested to apply the flag to both dict and table output as well.

So the way I see it right now you would want something like this:

Compact: node name by default visible, can be ommited with the node flag
Dict: node name by default ommited, can be made visible with the node flag
Table: node name by default ommited, can be made visible with the node flag
Json: node name always visible

Please correct me if I'm wrong here

cmd/observe/observe.go
```diff
@@ -242,6 +243,8 @@ programs attached to endpoints and devices. This includes:
 		"Translate IP addresses to logical names such as pod name, FQDN, ...",
 	)
 
+	observerCmd.Flags().BoolVarP(&node, "node", "", false, "Print node name in Compact output")
```
kAworu, line 246: nit: We already use "plain" names to define filters like `--pod`, `--namespace`, `--ip`. For consistency's sake I would rename this flag to something more explicit like `--display-node-name`.

kAworu (commented):
Thanks for the PR! Overall LGTM, though I agree with @glibsm suggestion.

Rolinh:
For reference: this PR addresses #86. It seems that displaying the node name is not what we want to do by default for a mode named "compact".
I'd be in favor of more uniformity at the cost of a small breaking change (that we can still afford pre-1.0).

What would make sense to me is to have the following:

- compact+dict+table: node name hidden by default, can be displayed with `--print-node-name` or whatever the flag ends up being called
- json: always print all fields (as is the case now)

I think having the flag toggling with different defaults per output type as suggested by @mdnix would be confusing to users.

mdnix:
> For reference: this PR addresses #86. It seems that displaying the node name is not what we want to do by default for a mode named "compact".
> I'd be in favor of more uniformity at the cost of a small breaking change (that we can still afford pre-1.0).
> 
> What would make sense to me is to have the following:
> 
> * compact+dict+table: node name hidden by default, can be displayed with `--print-node-name` or whatever the flag ends up being called
> * json: always print all fields (as is the case now)
> 
> I think having the flag toggling with different defaults per output type as suggested by @mdnix would be confusing to users.

Ok perfect. If this is alright with everyone I will reiterate and implement the change to match these criterias
Thanks for the feedback guys!

glibsm:
Sounds good. Lets do a small change before 1.0 because in hosted environments it's quite verbose

cmd/observe/observe.go
```diff
@@ -242,6 +243,8 @@ programs attached to endpoints and devices. This includes:
 		"Translate IP addresses to logical names such as pod name, FQDN, ...",
 	)
 
+	observerCmd.Flags().BoolVarP(&nodeName, "display-node-name", "", false, "Display node name in output")
```
Rolinh, line 246: nit: `print-node-name` is a bit shorter.

cmd/observe/observe.go
```diff
@@ -280,12 +292,20 @@ func handleArgs(ofilter *observeFilter) (err error) {
...
 	case "":
 		// no format specified, choose most appropriate format based on
 		// user provided flags
 		if follow {
-			opts = append(opts, hubprinter.Compact())
+			if nodeName {
+				opts = append(opts, hubprinter.Compact(), hubprinter.WithNodeName())
+			} else {
+				opts = append(opts, hubprinter.Compact())
+			}
```
Rolinh, line 308: I think the option could be added only once instead of individually for every output type, right? (assuming it does not matter to set it for the `json` output)
```go
if nodeName {
    opts = append(opts, hubprinter.WithNodeName())
} 
```

pkg/printer/printer_test.go
```diff
@@ -89,8 +89,8 @@ Jan  1 00:20:34.567   1.1.1.1:31793   2.2.2.2:8080   Policy denied   DROPPED   T
 				f: &f,
 			},
 			wantErr: false,
-			expected: "Jan  1 00:20:34.567 " +
-				"[k8s1]: 1.1.1.1:31793 -> 2.2.2.2:8080 " +
+			expected: "Jan  1 00:20:34.567: " +
+				"1.1.1.1:31793 -> 2.2.2.2:8080 " +
```
Rolinh, line 93: It would be nice to add a test case for the option with the node name.

pkg/printer/printer.go
```diff
@@ -194,65 +194,123 @@ func (p *Printer) WriteProtoFlow(res *observerpb.GetFlowsResponse) error {
...
+			_, err = fmt.Fprint(p.tw,
+				fmtTimestamp(f.GetTime()), tab,
+				f.GetNodeName(), tab,
+				src, tab,
+				dst, tab,
+				GetFlowType(f), tab,
+				f.GetVerdict().String(), tab,
+				f.GetSummary(), newline,
+			)
+		}
```
Rolinh, line 247: I'm not really fond of all this duplication just to add a new field when printing. This could lead to unwanted discrepancies based on whether the node name flag is active or not. What about something like this instead?
```go
func (p *Printer) WriteProtoFlow(res *observerpb.GetFlowsResponse) error {
        var err error
        write := func(w io.Writer, a ...interface{}) {
                if err != nil {
                        return
                }
                _, err = fmt.Fprint(w, a...)
        }

        f := res.GetFlow()
        switch p.opts.output {
        case TabOutput:
                var err error
                src, dst := p.GetHostNames(f)
                if p.line == 0 {
                        write(p.tw, "TIMESTAMP", tab)
                        if p.opts.nodeName {
                                write(p.tw, "NODE", tab)
                        }
                        write(p.tw,
                                "SOURCE", tab,
                                "DESTINATION", tab,
                                "TYPE", tab,
                                "VERDICT", tab,
                                "SUMMARY", newline,
                        )
                }
                write(p.tw, fmtTimestamp(f.GetTime()), tab)
                if p.opts.nodeName {
                        write(p.tw, f.GetNodeName(), tab)
                }
                write(p.tw,
                        src, tab,
                        dst, tab,
                        GetFlowType(f), tab,
                        f.GetVerdict().String(), tab,
                        f.GetSummary(), newline,
                )
                if err != nil {
                        return fmt.Errorf("failed to write out packet: %v", err)
                }
        case DictOutput:
            // apply the same method as for TabOutput, etc
            ...
```

Rolinh (commented):
I think this is OK but the code can be simplified to avoid duplication that could lead to bugs in the future.
Please, see suggestions.

pkg/printer/printer.go
```diff
@@ -242,17 +260,30 @@ func (p *Printer) WriteProtoFlow(res *observerpb.GetFlowsResponse) error {
...
-			fmtTimestamp(f.GetTime()),
-			f.GetNodeName(),
-			src,
-			dst,
-			GetFlowType(f),
-			f.GetVerdict().String(),
-			f.GetSummary(),
-		)
+
+		if !p.opts.nodeName {
```
glibsm, line 266: this can be one statement i think. we can make the whole `[node-name-here]` a value, and then pass in `""` in the no node name case

mdnix: That's a good idea. I will change it sometime tonight

pkg/printer/printer.go
```diff
@@ -220,18 +235,21 @@ func (p *Printer) WriteProtoFlow(res *observerpb.GetFlowsResponse) error {
...
+
 		if p.line != 0 {
 			// TODO: line length?
 			_, err := fmt.Fprintln(p.opts.w, dictSeparator)
 			if err != nil {
 				return err
 			}
 		}
-		src, dst := p.GetHostNames(f)
-		// this is a little crude, but will do for now. should probably find the
```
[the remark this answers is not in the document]
glibsm: three-part answer:

1. it's about manually formatting the `"   KEY:"` headers
2. yes it's still relevant
3. it's my own comment 🤦‍♂️

pkg/printer/printer.go
```diff
@@ -191,25 +192,39 @@ func GetFlowType(f v1.Flow) string {
...
+			return
+		}
+		_, err = fmt.Fprint(w, a...)
+	}
+
 	f := res.GetFlow()
+
 	switch p.opts.output {
 	case TabOutput:
+		var err error
```
kAworu, line 207: This is shadowing `err` declared at line 195. Because of that, the `err` check at line 234 will always take the false branch.

pkg/printer/printer.go
```diff
@@ -220,18 +235,21 @@ func (p *Printer) WriteProtoFlow(res *observerpb.GetFlowsResponse) error {
 			return fmt.Errorf("failed to write out packet: %v", err)
 		}
 	case DictOutput:
+		var err error
```
kAworu, line 238: This is shadowing `err` declared at line 195. Because of that, the `err` check at line 259 will always take the false branch.

Note that this `err` is actually used by line 243.

pkg/printer/printer.go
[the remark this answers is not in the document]
kAworu: Since it is still relevant, it should not be removed.

pkg/printer/printer.go
```diff
@@ -242,17 +260,22 @@ func (p *Printer) WriteProtoFlow(res *observerpb.GetFlowsResponse) error {
 			return fmt.Errorf("failed to write out packet: %v", err)
 		}
 	case CompactOutput:
+		var err error
```
kAworu, line 263: This is shadowing `err` declared at line 195. Although as it stand it is harmless, it is also useless.

kAworu (changes requested):
@mdnix I know the part of the patch I have concern about has been proposed by @Rolinh, but I'm not fan of the design and believe it has issues as it stand.

@gandro @glibsm I would be glad to have your feedback on this.

pkg/printer/printer.go
```diff
@@ -191,25 +192,39 @@ func GetFlowType(f v1.Flow) string {
 
 // WriteProtoFlow writes v1.Flow into the output writer.
 func (p *Printer) WriteProtoFlow(res *observerpb.GetFlowsResponse) error {
+	var err error
+	write := func(w io.Writer, a ...interface{}) {
+		if err != nil {
+			return
+		}
+		_, err = fmt.Fprint(w, a...)
+	}
```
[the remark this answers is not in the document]
gandro: For what it's worth, I'm fine with the code as is. Yes, the printer package could use a refactor to make error handling more pleasant, but that's out of scope for this PR. Let's not have a discussion about our error handling paradigms in Go in a PR which only attempts to add a simple feature.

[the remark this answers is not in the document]
Rolinh: > Checking for err before setting it is odd. After a careful read, I think I understand the idea: once err is not nil then write becomes a no-op, and we only have to check and report the error one we're done invoking write. This is bit too "cute" for my taste, and I would expect at least a comment.

FWIW, this is not my own invention but a pattern described by Rob Pike in this [blog post from 2015](https://blog.golang.org/errors-are-values) to avoid repetitive error-checking and I think this applies quite nicely here as it allows removing a whole bunch of error-checking while keeping the code readable.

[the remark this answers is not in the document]
kAworu: Thanks for the link!

mdnix:
I have implemented the suggestions of @kAworu  and added the errWriter suggestion of @Rolinh.

Rolinh (approved):
LGTM and the error is now properly scoped :100:

kAworu (approved):
LGTM now, @mdnix thanks for your contribution!

mdnix:
Thanks guys!

PR #362, opened 2020-09-16
Add Token to CMD Options

nathanjsweet:
For 3rd party auth providers that might intermediate
the request to hubble-relay.

Signed-off-by: Nate Sweet <nathanjsweet@pm.me>

### DIFF
--- a/cmd/observe/observe.go
+++ b/cmd/observe/observe.go
@@ -57,6 +57,7 @@ var (
 
 	debug   bool
 	numeric bool
+	token   string
 )
 
 func eventTypes() (l []string) {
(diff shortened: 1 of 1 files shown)


cmd/observe/observe.go
```diff
@@ -234,6 +235,12 @@ programs attached to endpoints and devices. This includes:
 		false,
 		"Display all information in numeric form",
 	)
+	observerCmd.Flags().StringVar(
+		&token,
+		"token",
```
glibsm, line 240: `token` seems rather generic for a CLI usage, we may want to add another word. `<some-qualifier>-token`

cmd/observe/observe.go
```diff
@@ -308,10 +315,26 @@ func handleArgs(ofilter *observeFilter) (err error) {
 	return nil
 }
 
+type creds string
+
+func (c creds) GetRequestMetadata(ctx context.Context, uri ...string) (map[string]string, error) {
+	return map[string]string{
+		"authorization": string(c),
```
michi-covalent, line 322: does this need `Bearer ` prefix?

nathanjsweet: No.

PR #369, opened 2020-09-29
cmd: remove globals, optimize grpc client conn creation, remove pprof

Rolinh:
This PR cleans up the `cmd` package:

- Remove pprof support. This was disabled by default in #269. AFAIK, it has never been used since then and is barely relevant anymore since the CLI is now a thin client to a Hubble server instance.
- Remove globals, ie no more global viper or command nor init func
- Cleanup gRPC client conn creation process.

See each commit for more details.

Note: this PR also introduces support for 2 new env variables:

- `HUBBLE_SERVER`: set the Hubble server address. The old `HUBBLE_DEFAULT_SOCKET_PATH` still works for compatibility reasons but this new env variable overrides it when provided.
- `HUBBLE_TIMEOUT`: set the dial timeout


gandro (approved):
Looks good!

PR #371, opened 2020-10-07
cmd/observe: Add HTTP method and path filters

twpayne:
This PR is a work-in-progress.

Adding the HTTP path filter is waiting on https://github.com/cilium/cilium/pull/13425 being merged into Cilium `master`.


gandro (approved):

cmd/observe/observe.go
```diff
@@ -179,6 +179,12 @@ programs attached to endpoints and devices. This includes:
 	observerCmd.Flags().Var(filterVar(
 		"http-status", ofilter,
 		"Show only flows which match this HTTP status code prefix (e.g. \"404\", \"5+\")"))
+	observerCmd.Flags().Var(filterVar(
+		"http-method", ofilter,
+		"Show only flows which match this HTTP method (e.g. \"get\", \"post\")"))
```
glibsm, line 184: Go has backtick to help with strings so you don't need to escape quotes

```
`Show only flows which match this HTTP method (e.g. "get", "post")"))`
```

cmd/observe/observe_filter.go
```diff
@@ -339,10 +341,29 @@ func (of *observeFilter) set(f *filterTracker, name, val string, track bool) err
 			})
 		}
 		f.apply(func(f *pb.FlowFilter) {
-			// TODO: auto-check L7?
 			f.HttpStatusCode = append(f.HttpStatusCode, val)
 		})
 
+	case "http-method":
+		if wipe {
```
glibsm, line 348: Someone can correct me, but I don't think `wipe` is relevant to `default nil` use-cases of filters right?. 

Perhaps we can look into rewriting this a bit so it's clearer when to use `wipe`.

gandro: Correct! I assume @twpayne copied this over from `http-status`, which also pontelessly wipes the value. This is not needed in any of the `http-*` flags.

https://github.com/cilium/hubble/blob/903e3c8b9b03aae88c331900c5cb45e4cd1fdfc7/cmd/observe/observe_filter.go#L40-L42

@twpayne Can you remove all the `wipe` from all `http-*` flags? It's technically an unrelated cleanup to this PR, but will make it a bit cleaner.

cmd/observe/observe.go
twpayne: Done. I also updated the other uses of double quotes.

cmd/observe/observe_filter.go
twpayne: Done.

gandro (approved):
Awesome! Thanks a lot for the additional cleanups as well!

PR #372, opened 2020-10-07
cmd: add support for TLS and mTLS

Rolinh:
Add a set of new global flags to configure the gRPC connection for mTLS
or TLS.

As part of this, 2 packages are added `cmd/common/{conn,validate}'. The
former implements a method to create a new gRPC connection according to
the global flags/config. The latter implement validation functions that
ensure consistency for a given set of flags/config.

There are two ways to specify that a connection must be established over
TLS:

- Use the `--tls` flag.
- The server address scheme starts with `tls://`.

Note that when a client certificate is provided to establish a
connection over mTLS; if neither of the above is set, the client refuses
to establish a connection and errors out mentioning that TLS is
required.

Example usage:
```
$ HUBBLE_SERVER=tls://localhost:4245 hubble status --tls-ca-cert-files ca.crt --tls-server-name "some.hubble-relay.cilium.io"
Current/Max Flows: 16384/16384 (100.00%)
Flows/s: 129.26
Connected Nodes: 4/4

$ HUBBLE_SERVER=tls://localhost:4245 hubble status --tls-ca-cert-files ca.crt --tls-allow-insecure
Current/Max Flows: 16384/16384 (100.00%)
Flows/s: 129.53
Connected Nodes: 4/4

$ HUBBLE_SERVER=localhost:4245 hubble status --tls-ca-cert-files ca.crt --tls-allow-insecure --tls
Current/Max Flows: 16384/16384 (100.00%)
Flows/s: 130.14
Connected Nodes: 4/4

$ HUBBLE_SERVER=localhost:4245 ./hubble status --tls-ca-cert-files ca.crt --tls-server-name "some.hubble-relay.cilium.io" --tls-client-cert-file client.crt
invalid flag(s): certificate and private key are both required, but only one was provided

$HUBBLE_SERVER=localhost:4245 ./hubble status --tls-ca-cert-files ca.crt --tls-server-name "some.hubble-relay.cilium.io" --tls-client-cert-file client.crt --tls-client-key-file client.key 
invalid flag(s): transport layer security required
```

Note that these new flags may also be provided as environment variable (eg `HUBBLE_TLS_SERVER_NAME="some.hubble-relay.cilium.io"`).

As an added bonus, this PR also improves error messages on failed dialing attempts (requires a newer grpc-go version, hence the vendor bump). So instead of a close-to-useless `context deadline exceeded`, we now get errors like:
```
$ HUBBLE_SERVER=tls://localhost:4245 hubble status
failed to connect to 'tls://localhost:4245': context deadline exceeded: connection error: desc = "transport: authentication handshake failed: x509: certificate is valid for *.hubble-relay.cilium.io, not localhost"

$ HUBBLE_SERVER=tls://localhost:4245 hubble status 
failed to connect to 'tls://localhost:4245': connection error: desc = "transport: error while dialing: dial tcp [::1]:4245: connect: connection refused"
```


cmd/common/validate/tls.go
```diff
@@ -0,0 +1,57 @@
...
+	switch {
+	case vp.GetString("tls-client-key-file") != "" && vp.GetString("tls-client-cert-file") != "":
+		needTLS = true
+	case vp.GetString("tls-client-key-file") != "":
+		fallthrough
+	case vp.GetString("tls-client-cert-file") != "":
+		return ErrInvalidKeypair
+	}
+
+	if needTLS && !(vp.GetBool("tls") || strings.HasPrefix(vp.GetString("server"), defaults.TargetTLSPrefix)) {
```
gandro, line 53: I think (at least according to what we discussed in the last SIG-Hubble meeting) that we want to automatically imply `--tls` if the user specifies credentials such as a client certificate, instead of throwing an error here.

cmd/root.go
```diff
@@ -53,14 +63,50 @@ func New() *cobra.Command {
...
+	flags.String(
+		"tls-server-name",
+		"",
+		"Specify a server name to verify the hostname on the returned certificate (eg: 'instance.hubble-relay.cilium.io').",
+	)
 	vp.BindPFlags(flags)
 
 	rootCmd.SetErr(os.Stderr)
-	rootCmd.SetVersionTemplate("{{with .Name}}{{printf \"%s \" .}}{{end}}{{printf \"v%s\" .Version}}\n")
+	rootCmd.SetVersionTemplate("{{with .Name}}{{printf \"%s \" .}}{{end}}{{printf \"v%s\" .Version}}\r\n")
```
gandro, line 109: What's the rationale for this? Windows support? I did not see any comment in the commit message.

cmd/root.go
```diff
@@ -53,14 +63,50 @@ func New() *cobra.Command {
...
-	flags.String("config", "", "Config file (default is $HOME/.hubble.yaml)")
-	flags.BoolP("debug", "D", false, "Enable debug messages")
-	flags.String("server", defaults.GetDefaultSocketPath(), "Address of a Hubble server")
-	flags.Duration("timeout", defaults.DefaultDialTimeout, "Hubble server dialing timeout")
+	flags.String("config", "", "Config file (default is $HOME/.hubble.yaml).")
+	flags.BoolP("debug", "D", false, "Enable debug messages.")
+	flags.String("server", defaults.GetDefaultSocketPath(), "Address of a Hubble server.")
+	flags.Duration("timeout", defaults.DefaultDialTimeout, "Hubble server dialing timeout.")
+	flags.Bool(
+		"tls",
```
gandro, line 71: nit: Maybe we want to name this `--tls-required` and (as mentioned above) implicitly enable TLS whenever one of the following applies:

- The scheme starts with `tls://`
- The user has provided a client certificate
- (maybe?) The user has provided any of the other TLS flags (servername, ca)

The semantics of the `tls-required` would be:
- `false`: Enable TLS if one of the above applies
- `true`: Enable TLS unconditionally

gandro (changes requested):
Nice! I'm leaving a blocking review since I think there are some semantic questions that need to be resolved first, but the overall code looks very solid to me!

cmd/root.go
Rolinh: Oh right, just Windows support refinement :grin:

cmd/common/validate/tls.go
Rolinh: I'm really not fond of CLIs that try to be too smart and automagically change their behavior based on a combination of specific user input. This is why I opted for to add the `tls://` scheme (which I think we did not consider during the last SIG-Hubble meeting) so that the CLI behavior is really predictable. Let's maybe discuss this during today's SIG-Hubble meeting.

gandro (approved):
As discussed in SIG-Hubble, we will not automagically enable TLS

cmd/common/conn/tls.go
```diff
@@ -0,0 +1,74 @@
...
+	// optional mTLS
+	clientCertFile := vp.GetString("tls-client-cert-file")
+	clientKeyFile := vp.GetString("tls-client-key-file")
+	if clientCertFile != "" && clientKeyFile != "" {
+		cert, err := tls.LoadX509KeyPair(clientCertFile, clientKeyFile)
+		if err != nil {
+			return nil, err
+		}
+		tlsConfig.Certificates = []tls.Certificate{cert}
+	}
```
kAworu, line 70: IMO it would be nice to use Cilium's `pkg/crypto/certloader` stuff here so that we won't have to implement the `x509.NewCertPool()` and `tls.LoadX509KeyPair()` dance all over again.

At the moment, Cilium's certloader stuff is always running a file watcher which in the context of Hubble CLI is overkill. Maybe it would be a good idea to extract the certloader in its own go module / repository so that we can extend it with what CLI needs.

In the mean time (or in any case), this PR could port [how the certloader client is handling the keypair using the `GetClientCertificate` callback rather than the `tlsConfig.Certificates` directly](https://github.com/cilium/cilium/blob/master/pkg/crypto/certloader/client.go#L66-L69). The value here is when the server is configured for mTLS but a cli invocation fail to provide the client keypair. This is what we get with the patch as it stand:

```
% ./hubble observe \
    --server tls://localhost:4244 \
    --tls-ca-cert-files ca.pem \
    --tls-server-name foo.default.hubble-grpc.cilium.io
failed to connect to 'tls://localhost:4244': context deadline exceeded: connection closed
```

And this is what we could get using the `GetClientCertificate` callback:

```
% ./hubble observe \
    --server tls://localhost:4244 \
    --tls-ca-cert-files ca.pem \
    --tls-server-name foo.default.hubble-grpc.cilium.io
failed to connect to 'tls://localhost:4244': context deadline exceeded: connection error: desc = "transport: authentication handshake failed: mTLS client certificate requested, but not configured"
```

kAworu (changes requested):
Great diff! Small improvement suggested and nit.

cmd/root.go
```diff
@@ -41,6 +43,14 @@ func New() *cobra.Command {
 		SilenceErrors: true, // this is being handled in main, no need to duplicate error messages
 		SilenceUsage:  true,
 		Version:       pkg.Version,
+		PersistentPreRunE: func(cmd *cobra.Command, _ []string) error {
+			for _, fn := range validate.FlagFuncs {
+				if err := fn(cmd, vp); err != nil {
+					return fmt.Errorf("invalid flag(s): %w", err)
+				}
+			}
```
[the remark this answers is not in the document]
Rolinh: Added `validate.Flags(cmd, vp)`.

kAworu (approved):
LGTM!

PR #375, opened 2020-10-08
cmd: honor user configuration directory for the configuration file

Rolinh:
Honor platform specific user configuration directory to look for a
configuration file for hubble. In addition, look up in the current
directory first and fallback to `$HOME/.hubble`.

On Unix systems, the user specific configuration directory is
`$XDG_CONFIG_HOME` if non-empty, else `$HOME/.config`. On MacOS, it's
`$HOME/Library/Application Support` and on Windows it's `%AppData%`.

For example, on Linux the configuration file is now looked for in this
order:

1. `$PWD/config.yaml`
2. `$XDG_CONFIG_HOME/hubble/config.yaml`
3. `$HOME/.config/hubble/config.yaml`
4. `$HOME/.hubble/config.yaml`

Note that this is a breaking change as previously the configuration file
was expected to be found in `$HOME/.hubble.yaml`. As the latter is not
standard and not recommended, let's opt for the freedesktop
recommendation and avoid polluting the user's home directory root with a
hidden config file.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -47,7 +47,7 @@ func New() *cobra.Command {
 			vp.SetConfigFile(cfg)
 		}
 		// if a config file is found, read it in.
-		if err := vp.ReadInConfig(); err == nil {
+		if err := vp.ReadInConfig(); err == nil && vp.GetBool("debug") {
 			fmt.Fprintln(rootCmd.ErrOrStderr(), "Using config file:", vp.ConfigFileUsed())
 		}
 	})
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -17,6 +17,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/cilium/hubble/cmd/completion"
 	"github.com/cilium/hubble/cmd/observe"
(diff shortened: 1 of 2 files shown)


kAworu (approved):
LGTM.

Question: is this really a breaking change? We still fallback to home directory correct?

kAworu (approved):

cmd/root.go
```diff
@@ -80,11 +81,18 @@ func Execute() error {
...
-	vp.AutomaticEnv()           // read in environment variables that match
+
+	// read config from a file
+	vp.SetConfigName("config") // name of config file (without extension)
+	vp.SetConfigType("yaml")   // useful if the given config file does not have the extension in the name
+	vp.AddConfigPath(".")      // look for a config in the working directory first
+	if dir, err := os.UserConfigDir(); err == nil {
+		vp.AddConfigPath(filepath.Join(dir, "hubble")) // honor user config dir
+	}
+	vp.AddConfigPath(filepath.Join("$HOME", ".hubble")) // fallback to home directory
```
gandro, line 92: nit: 
```suggestion
if dir, err := os.UserHomeDir(); err == nil {
	vp.AddConfigPath(filepath.Join(dir, ".hubble")) // fallback to home directory
}
```

Rolinh: Good point!

PR #377, opened 2020-10-08
cmd: fix help message for the `-config` flag

Rolinh:
The configuration file location is now platform dependent.


kAworu (approved):

Rolinh:
@kAworu I actually figured that we can be more precise and specify where to config file should be placed in the help message. PTAL.

kAworu (approved):
:100:

PR #378, opened 2020-10-08
Add reflect command

michi-covalent:
This command sends gRPC server reflection request to Hubble server
and prints out the response as JSON. It's marked hidden since it's
mainly used during development/troubleshooting.

    % hubble reflect | jq
    {
      "name": "observer/observer.proto",
      "package": "observer",
      "dependency": [
        "google/protobuf/wrappers.proto",
        "google/protobuf/timestamp.proto",
        "flow/flow.proto",
        "relay/relay.proto"
      ],
      ...
    }


cmd/reflect/reflect.go
```diff
@@ -0,0 +1,160 @@
...
+func New() *cobra.Command {
+	reflectCmd := &cobra.Command{
+		Use:   "reflect",
+		Short: "Let Hubble reflect upon itself",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runReflect()
+		},
+		Hidden: true,
+	}
+	reflectCmd.Flags().StringVarP(&serverURL, "server", "", defaults.GetDefaultSocketPath(), "URL to connect to server")
```
Rolinh, line 45: Since #369, `server` is a global persistent flag that can also be set via an environment variable (`HUBBLE_SERVER`).
Similarly to `peer.New`, `observe.New` and `status.New`, `reflect.New` should take a viper as an argument (`New(vp *viper.Viper`) and use `vp.GetString("server")` to retrieve the server URL.

cmd/reflect/reflect.go
```diff
@@ -0,0 +1,160 @@
...
+func runReflect() (err error) {
+	conn, err := grpc.Dial(serverURL, grpc.WithInsecure())
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if closeErr := conn.Close(); closeErr != nil && err == nil {
+			err = closeErr
+		}
+	}()
```
Rolinh, line 58: Since #369, we don't duplicate code to create a new `conn` and use `cmd/common`.  PR #372 changes this a bit as well to ensure consistent settings are passed when creating a new `conn`. 

The TLDR is that you should init the conn in `RunE` and pass it down to `runReflect`. It would look something like this:
```go
import "github.com/cilium/hubble/cmd/common/conn"

func New(vp *viper.Viper) *cobra.Command {
	return := &cobra.Command{
		Use:   "reflect",
		Short: "Let Hubble reflect upon itself",
		RunE: func(_ *cobra.Command, _ []string) error {
                        ctx, cancel := context.WithCancel(context.Background())
                        defer cancel()
                        hubbleConn, err := conn.New(ctx, vp.GetString("server"), vp.GetDuration("timeout"))
                        if err != nil {
                                return err
                        }
                        defer hubbleConn.Close()
                        return runReflect(ctx, hubbleConn)
		},
		Hidden: true,
	}

```

cmd/reflect/reflect.go
```diff
@@ -0,0 +1,160 @@
...
+
+var (
+	serverURL string
+)
+
+// New returns the reflect command.
+func New() *cobra.Command {
+	reflectCmd := &cobra.Command{
+		Use:   "reflect",
+		Short: "Let Hubble reflect upon itself",
```
Rolinh, line 39: What about this?
```suggestion
		Short: "Use gRPC reflection to explore Hubble's API",
```

cmd/reflect/reflect.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	// same proto file can be imported multiple times from different places. keep track of proto filenames
+	// so that we don't print them multiple times.
+	visited := make(map[string]bool)
+	encoder := json.NewEncoder(os.Stdout)
+	for _, svc := range services.ListServicesResponse.Service {
+		if svc.Name == "grpc.reflection.v1alpha.ServerReflection" || svc.Name == "grpc.health.v1.Health" {
+			continue
+		}
+		err = client.Send(&rpb.ServerReflectionRequest{
```
Rolinh, line 90: It looks like `err` is not checked.

cmd/reflect/reflect.go
```diff
@@ -0,0 +1,160 @@
...
+		return err
+	}
+	services, ok := res.GetMessageResponse().(*rpb.ServerReflectionResponse_ListServicesResponse)
+	if !ok {
+		return fmt.Errorf("unexpected response: %v", res)
+	}
+	// same proto file can be imported multiple times from different places. keep track of proto filenames
+	// so that we don't print them multiple times.
+	visited := make(map[string]bool)
+	encoder := json.NewEncoder(os.Stdout)
```
Rolinh, line 85: Can we honor `cobra.OutOrStdout()` here?

Rolinh (changes requested):
Awesome, thanks for this it'll be incredibly useful to explore the API when working on API changes!! :tada: 

There are a couple of changes required though, please see comments. I'll also mark the PR is blocked by #372 as the grpc client conn creation logic is refactored.

cmd/reflect/reflect.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		return err
+	}
+	services, ok := res.GetMessageResponse().(*rpb.ServerReflectionResponse_ListServicesResponse)
+	if !ok {
+		return fmt.Errorf("unexpected response: %v", res)
+	}
+	// same proto file can be imported multiple times from different places. keep track of proto filenames
+	// so that we don't print them multiple times.
+	visited := make(map[string]bool)
```
glibsm, line 84: `map[string]struct{}` is technically more efficient

glibsm (commented):
👏 nice

cmd/reflect/reflect.go
michi-covalent: yeah i always forget about this 🤯

Rolinh (approved):

kAworu (approved):
:rocket:

PR #380, opened 2020-10-12
Add config subcommand

Rolinh:
```
$ hubble config -h
Modify or view hubble config

Usage:
  hubble config [flags]
  hubble config [command]

Available Commands:
  get         Get an individual value in the hubble config file
  reset       Reset all or an individual value in the hubble config file
  set         Set an individual value in the hubble config file
  view        Display merged configuration settings
...
```

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -1,0 +1,39 @@
+// Copyright 2020 Authors of Hubble
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+// New config command.
+func New(vp *viper.Viper) *cobra.Command {
+	configCmd := &cobra.Command{
+		Use:   "config",
+		Short: "Modify or view hubble config",
+		Long:  "Modify or view hubble config",
+		PersistentPreRunE: func(_ *cobra.Command, _ []string) error {
+			// override root persistent pre-run to avoid flag/config checks
+			// as we want to be able to modify/view the config even if it is
+			// invalid
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			return cmd.Help()
+		},
+	}
+	return configCmd
+}
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -36,7 +36,17 @@ func New(vp *viper.Viper) *cobra.Command {
 		},
 	}
 	configCmd.AddCommand(
+		newSetCommand(vp),
 		newViewCommand(vp),
 	)
 	return configCmd
+}
+
+func isKey(vp *viper.Viper, key string) bool {
+	for _, k := range vp.AllKeys() {
+		if key == k {
+			return true
+		}
+	}
+	return false
 }
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -36,6 +36,7 @@ func New(vp *viper.Viper) *cobra.Command {
 		},
 	}
 	configCmd.AddCommand(
+		newResetCommand(vp),
 		newSetCommand(vp),
 		newViewCommand(vp),
 	)
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -36,6 +36,7 @@ func New(vp *viper.Viper) *cobra.Command {
 		},
 	}
 	configCmd.AddCommand(
+		newGetCommand(vp),
 		newResetCommand(vp),
 		newSetCommand(vp),
 		newViewCommand(vp),
--- a/cmd/config/get.go
+++ b/cmd/config/get.go
@@ -1,0 +1,76 @@
+// Copyright 2020 Authors of Hubble
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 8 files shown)


cmd/config/config.go
```diff
@@ -0,0 +1,37 @@
...
+// limitations under the License.
+
+package config
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+// New status command.
```
glibsm, line 22: s/status/config

cmd/config/view.go
```diff
@@ -0,0 +1,43 @@
...
+		},
+	}
+}
+
+func runView(cmd *cobra.Command, vp *viper.Viper) error {
+	bs, err := yaml.Marshal(vp.AllSettings())
+	if err != nil {
+		return fmt.Errorf("failed to marshal config to YAML: %w", err)
+	}
+	fmt.Fprint(cmd.OutOrStdout(), string(bs))
```
glibsm, line 41: I think without `Fprintln` your shell won't properly show the new prompt

cmd/config/config.go
```diff
@@ -0,0 +1,37 @@
...
+		Use:   "config",
+		Short: "Modify or view hubble config",
+		Long:  "Modify or view hubble config",
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			return cmd.Help()
+		},
+	}
+	configCmd.AddCommand(
+		newSetCommand(vp),
+		newViewCommand(vp),
```
glibsm, line 34: Perhaps `hubble config get <KEY>` also?

Or meybe an optional `hubble config view <KEY>`.

Not blocking since easily mitigated via jq or grep.

cmd/config/set.go
```diff
@@ -0,0 +1,96 @@
...
+			for _, s := range strings.Split(value, ",") {
+				ss = append(ss, strings.TrimSpace(s))
+			}
+			newVal = ss
+		default:
+			return fmt.Errorf("unhandeld kind %s of %s, please open an issue", v.Kind(), iv.Kind())
+		}
+	case reflect.String:
+		newVal = value
+		// FIXME: find a way to figure out which keys are of type duration
```
glibsm, line 83: 🤔

Rolinh: Values of type `time.Duration` are stored as string (eg:  "3s"). If we don't do this check, a user can `hubble config set timeout foobar`, which works, but then the value is unusable. Not sure of a good way to deal with this atm.

glibsm: Yeah I understand the problem. I suppose there is no harm to display them as-is, right? If timeout is set to `5h` displaying that string shouldn't be confusing to the user.

Why do we have to cast necessarily?

cmd/config/view.go
Rolinh: I actually get an extra newline when using `Fprintln` here.

cmd/config/set.go
glibsm: Deleted my previous comment, because I misunderstood the function.

Did you lift this function somewhere from cobra-viper integration, or did you write it from scratch? Somewhere in the cobra-viper path there should be something very similar on casting strings to their correct types.

cmd/config/view.go
glibsm: Huh. I suppose yaml already marshals pretty printed

cmd/config/set.go
Rolinh: I wrote it from scratch. As we bind viper to cobra flags, I figured out a more robust way to assign the correct type for the value. As an added bonus, one can now omit providing a value which will reset it to its default. This new capability also prompted me to add `hubble config reset`.

cmd/config/config.go
Rolinh: Added get command.

cmd/config/set.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	val := value
+	if value == "" {
+		val = flag.DefValue
+	}
+
+	var err error
+	var newVal interface{}
+	typ := flag.Value.Type()
```
glibsm, line 66: Looks like viper also has `flag.ValueType()`. Unclear if it's any different than what you wrote

https://github.com/cilium/hubble/blob/master/vendor/github.com/spf13/viper/viper.go#L1053

cmd/config/get.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	// each viper key/val entry should be bound to a command flag
+	flag := cmd.Flag(key)
+	if flag == nil {
+		return fmt.Errorf("key=%s not bound to a flag", key)
+	}
+
+	var val interface{}
+	switch typ := flag.Value.Type(); typ {
```
glibsm, line 57: Viper uses a similar get, but without string versions of the types

https://github.com/cilium/hubble/blob/master/vendor/github.com/spf13/viper/viper.go#L742

glibsm (approved):

cmd/config/set.go
Rolinh: It looks like it's [literally](https://github.com/cilium/hubble/blob/2b8c000824cdfc9ff9fb64d9bac408de795f9c61/vendor/github.com/spf13/viper/flags.go#L54-L57) the same:
```go
// ValueType returns the type of the flag as a string.
func (p pflagValue) ValueType() string {
	return p.flag.Value.Type()
}
```

gandro (changes requested):
I did test this without reading to much about how it is implemented and I find this behavior here a bit surprising. Is it expected or a bug?

```console
$ ./hubble config view
config: /home/gandro/.config/hubble/config.yaml
debug: false
server: unix:///var/run/cilium/hubble.sock
timeout: 5s
tls: false
tls-allow-insecure: false
tls-ca-cert-files: []
tls-client-cert-file: ""
tls-client-key-file: ""
tls-server-name: ""
$ ./hubble config set timeout 4s
unknown key: timeout
```

Rolinh:
@gandro I'm unable to reproduce:
```
$ ./hubble config view
config: /home/isovalent/.config/hubble/config.yaml
debug: false
server: localhost:4245
timeout: 3s
tls: false
tls-allow-insecure: false
tls-ca-cert-files: []
tls-client-cert-file: ""
tls-client-key-file: ""
tls-server-name: ""
$ ./hubble config set timeout 4s
$ ./hubble config view
config: /home/isovalent/.config/hubble/config.yaml
debug: false
server: localhost:4245
timeout: 4s
tls: false
tls-allow-insecure: false
tls-ca-cert-files: []
tls-client-cert-file: ""
tls-client-key-file: ""
tls-server-name: ""
```

gandro (approved):
Confirmed that this now works if the file or parts of the file are absent.

PR #381, opened 2020-10-12
vendor: go mod tidy && go mod vendor && go mod verify

Rolinh:
It looks like we forgot to tidy the vendor dir in #378.

PR #382, opened 2020-10-12
actions: add go-mod check

Rolinh:
This check ensures that vendored deps are "clean".

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -32,3 +32,18 @@ jobs:
       with:
         name: release
         path: release
+  go-mod:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Go
+        uses: actions/setup-go@v1
+        with:
+          go-version: '1.15.2'
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Check module vendoring
+        run: |
+          go mod tidy
+          go mod vendor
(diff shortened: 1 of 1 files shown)


Rolinh:
Blocked on #381

glibsm (approved):

michi-covalent (approved):
thanks @Rolinh for me `go mod tidy` was failing with this error:
```
% go mod tidy
go: google.golang.org/grpc@v1.33.0: reading google.golang.org/grpc/go.mod at revision v1.33.0: unknown revision v1.33.0
```
and i didn't catch the error.

.github/workflows/unittests.yml
```diff
@@ -32,3 +32,18 @@ jobs:
       with:
         name: release
         path: release
+  go-mod:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Go
+        uses: actions/setup-go@v1
+        with:
+          go-version: '1.15.2'
```
michi-covalent, line 41: should we bump go version in go.mod as well? https://github.com/cilium/hubble/blob/f7d360b8921a5b958e07109983b8ca5e9cb40ab8/go.mod#L3

Rolinh: I don't think we have any reason to unless we want to use Go features that are not available in go 1.14.

PR #385, opened 2020-10-14
cmd/config: only write provided key/value when using set subcommand

Rolinh:
Address an issue where values other than the one specified using `config
set KEY ...` were set to their default value in the configuration file.
Key/value pairs provided via environment variables and/or flags were
also written to the file, which was an unwanted side-effect of the
previous implementation.

With this rewrite, only values that are explicitly set via the `config
set` subcommand are written to the configuration file. Key/value pairs
that were already in the configuration file, if any, are preserved.

### DIFF
--- a/cmd/config/set.go
+++ b/cmd/config/set.go
@@ -16,7 +16,9 @@ package config
 
 import (
 	"encoding/csv"
+	"errors"
 	"fmt"
+	"path/filepath"
 	"strings"
 
 	"github.com/spf13/cast"
@@ -50,7 +52,7 @@ func runSet(cmd *cobra.Command, vp *viper.Viper, key, value string) error {
 		return fmt.Errorf("unknown key: %s", key)
 	}
 
-	// each viper key/val entry should be bound to a command flag
+	// each viper key/value entry should be bound to a command flag
 	flag := cmd.Flag(key)
 	if flag == nil {
 		return fmt.Errorf("key=%s not bound to a flag", key)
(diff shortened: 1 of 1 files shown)


cmd/config/set.go
```diff
@@ -86,6 +88,43 @@ func runSet(cmd *cobra.Command, vp *viper.Viper, key, value string) error {
 	if err != nil {
 		return fmt.Errorf("cannot assign value=%s for key=%s, expected type=%s: %w", value, key, typ, err)
 	}
-	vp.Set(key, newVal)
-	return vp.WriteConfig()
+
+	// Create a file only viper from the configured file to avoid writing
```
gandro, line 92: nit (had difficulty parsing this the first time around):
```suggestion
	// Create a file-only viper config from the configured file to avoid writing
```

cmd/config/set.go
```diff
@@ -86,6 +88,43 @@ func runSet(cmd *cobra.Command, vp *viper.Viper, key, value string) error {
...
+	typ := strings.TrimPrefix(ext, ".")
+
+	vp := viper.New()
+	vp.SetConfigName(filename)
+	vp.SetConfigType(typ)
+	vp.AddConfigPath(dir)
+
+	// tentatively read the config file
+	// it will fail if the config file does not exist, which is fine
+	_ = vp.ReadInConfig()
```
gandro, line 128: I think there can be other errors (such as bad file format) in which case we do not silently want to ignore them, as it would cause use to accidentally overwrite the file I believe (e.g. in case someone specifies `--config somefile.txt` where somefile.txt is not a hubble config)

 Can we check for `viper.ConfigFileNotFoundError` as documented here: https://github.com/spf13/viper#reading-config-files

gandro (changes requested):

gandro (approved):

PR #391, opened 2020-10-18
vendor: bump cobra to v1.1.1

twpayne:
cobra 1.1.0 inadvertently included a breaking YAML change affecting
kubectl and helm. See https://github.com/spf13/cobra/pull/1259 for
details.

Signed-off-by: Tom Payne <tom@isovalent.com>


Rolinh (changes requested):
I think this won't be enough as our `go.mod` still points to `gopkg.in/yaml.v2 v2.3.0` (ie the version with the unintended breaking change that cobra v1.1.1 reverted). Please, also downgrade the `yaml` to `gopkg.in/yaml.v2 v2.2.8`.

sayboras:
> I think this won't be enough as our `go.mod` still points to `gopkg.in/yaml.v2 v2.3.0` (ie the version with the unintended breaking change that cobra v1.1.1 reverted). Please, also downgrade the `yaml` to `gopkg.in/yaml.v2 v2.2.8`.

Might need to use replace directive to enforce  v2.2.8 as well. Thinking about it, I noticed that cilium is using v2.3.0, is there anything to worry about? /cc @Rolinh @aanm

twpayne:
Thanks for the feedback, PR updated to pin `gopkg.in/yaml.v2` to `v2.2.8` as suggested by @sayboras.

Rolinh (approved):

sayboras (approved):
LGTM :100: 

Seems like I have permission in hubble as well, I thought it's only cilium repo :)

PR #392, opened 2020-10-19
vendor: bump cilium to master right before branching v1.9

Rolinh:
Note: I created a separate `replace` section for Hubble's own replace directives so it's easier to replace the ones from Cilium when updating.


gandro (approved):
Thanks!

twpayne (approved):

PR #396, opened 2020-10-19
release: fix `release` binary usage instruction

Rolinh:
This tool defaults to `cilium/cilium` as the base repository so one needs to pass `--repo cilium/hubble`.
Also, add a `v` in `--base v$LAST_RELEASE` as Github can't find the version to compare to otherwise.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -68,7 +68,7 @@ to reflect the next planned release, i.e.
 Using https://github.com/cilium/release, prepare the release notes between the
 last minor version (latest patch) and current.
 
-    ./release --base $LAST_RELEASE --head v$MAJOR.$MINOR
+    ./release --repo cilium/hubble --base v$LAST_RELEASE --head v$MAJOR.$MINOR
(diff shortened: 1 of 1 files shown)


gandro (approved):

RELEASE.md
```diff
@@ -68,7 +68,7 @@ to reflect the next planned release, i.e.
 Using https://github.com/cilium/release, prepare the release notes between the
 last minor version (latest patch) and current.
 
-    ./release --base $LAST_RELEASE --head v$MAJOR.$MINOR
+    ./release --repo cilium/hubble --base v$LAST_RELEASE --head v$MAJOR.$MINOR
```
glibsm, line 71: I think I wrote this to intend `v` to be part of the `$LAST_RELEASE`, as in the latest tag that is known on that branch (but not documented it properly). If someone puts a `v` in their last release it still won't work :)

Either way works, really.

glibsm (approved):

PR #406, opened 2020-10-21
doc: fix broken links

Rolinh:


gandro (approved):
Unrelated to the PR, we might want to switch to `stable` docs going forward. `latest` tend to not be always consistent.

PR #411, opened 2020-10-28
cmd/observe: add all flags

### LINKED ISSUE #253: Add observe all
`hubble observe --all` is more friendly to get the whole buffer rather than something like `hubble observe --since 100h`

Rolinh:
Closes #253

### DIFF
--- a/cmd/observe/observe.go
+++ b/cmd/observe/observe.go
@@ -40,6 +40,7 @@ import (
 
 var (
 	last               uint64
+	all                bool
 	sinceVar, untilVar string
 
 	jsonOutput          bool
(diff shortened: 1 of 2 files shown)


cmd/observe/observe.go
```diff
@@ -103,7 +104,10 @@ more.`,
 		"type", "t", ofilter, []string{},
 		fmt.Sprintf("Filter by event types TYPE[:SUBTYPE] (%v)", eventTypes())))
 
-	observerCmd.Flags().Uint64Var(&last, "last", 0, fmt.Sprintf("Get last N flows stored in the hubble (default %d)", defaults.FlowPrintCount))
+	observerCmd.Flags().Uint64Var(&last, "last", 0, fmt.Sprintf("Get last N flows stored in Hubble's buffer (default %d)", defaults.FlowPrintCount))
+	observerCmd.Flags().BoolVar(&all, "all", false, "Get all flows stored in Hubble's buffer")
+	observerCmd.Flags().BoolVar(&dumpAll, "dump-all", false, "Dump all flows without applying any filter (only useful for debugging)")
+	observerCmd.Flags().MarkHidden("dump-all") // this is only useful to developers for debugging purposes
```
glibsm, line 110: I don't understand the point of this, rather than calling `hubble observe --all`. What's the difference?

Rolinh: I updated the description in the Cilium PR to add more context around this: https://github.com/cilium/cilium/pull/13799#issue-511575570. There could be a way to implement `--all --follow` and avoid `ErrInvalidRead` errors but this could be more complex, would probably mean 2 concurrent readers are necessary and induce a slightly increased memory consumption. There is a case to be made about it because if implemented, it would obsolete `dump` altogether.

Rolinh: I put the cilium PR that adds support for `dump` in draft mode as I want to experiment with the 2 readers approach described here https://github.com/cilium/cilium/pull/13799#discussion_r514443394

Therefore, I have removed the commit that adds the `--dump` flag to this PR and only kept the one that adds `--all`.

glibsm (approved):

PR #420, opened 2020-11-06
cmd/config: add shell completion support for keys for get|set|reset

Rolinh:
Add support for shell completion for keys for the `get`, `set` and
`reset` subcommands. This means that when using `<TAB>` after
`config get|set|reset`, the list of possible configuration keys is
offered for completion, improving the user experience.

Example with the `get` subcommand (using `zsh`):

```
$ hubble config get <TAB>
config                server                tls                   tls-ca-cert-files     tls-client-key-file
debug                 timeout               tls-allow-insecure    tls-client-cert-file  tls-server-name
```


glibsm (approved):
Nice.

I think it's possible to write a simple test for this, since completion just spews a bunch of text out which can be validated.

PR #421, opened 2020-11-06
cmd/observe: add shell completion support for various flags

Rolinh:
Some of the flags possible values are known. This is valid for flags:
  - type
  - verdict
  - http-status
  - http-method
  - output

Example:

    hubble observe --type <TAB>
    agent           capture         debug           drop            l7              policy-verdict  trace

Adding completion support for other flags would require dynamic
completion and, in most cases, access to the kubernetes cluster (to
complete things like namespace, pod, service, etc) which makes it much
harder to implement.

### DIFF
--- a/cmd/observe/observe.go
+++ b/cmd/observe/observe.go
@@ -19,8 +19,10 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"os"
 	"os/signal"
+	"strings"
 
 	pb "github.com/cilium/cilium/api/v1/flow"
 	"github.com/cilium/cilium/api/v1/observer"
@@ -56,6 +58,12 @@ var (
 
 	numeric bool
 )
+
+var verdicts = []string{
+	pb.Verdict_FORWARDED.String(),
+	pb.Verdict_DROPPED.String(),
+	pb.Verdict_ERROR.String(),
+}
 
 func eventTypes() (l []string) {
 	for t := range monitorAPI.MessageTypeNames {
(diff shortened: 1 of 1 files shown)


cmd/observe/observe.go
```diff
@@ -177,14 +180,53 @@ more.`,
...
-		fmt.Sprintf("Show only flows with this verdict [%v, %v]", pb.Verdict_FORWARDED, pb.Verdict_DROPPED)))
+		fmt.Sprintf(
+			"Show only flows with this verdict [%v, %v, %v]",
+			pb.Verdict_FORWARDED, pb.Verdict_DROPPED, pb.Verdict_ERROR),
+	))
+	observerCmd.RegisterFlagCompletionFunc("verdict", func(_ *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
+		return []string{
+			pb.Verdict_FORWARDED.String(),
+			pb.Verdict_DROPPED.String(),
+			pb.Verdict_ERROR.String(),
```
glibsm, line 191: i think these two should use the same `[]string` slice, just one can do `strings.Join(",")`

cmd/observe/observe.go
```diff
@@ -177,14 +180,53 @@ more.`,
...
+			"200", "201", "202", "203", "204", "205", "206", "207", "208",
+			"226",
+			"300", "301", "302", "303", "304", "305", "307", "308",
+			"400", "401", "402", "403", "404", "405", "406", "407", "408", "409",
+			"410", "411", "412", "413", "414", "415", "416", "417", "418",
+			"421", "422", "423", "424", "425", "426", "428", "429",
+			"431",
+			"451",
+			"500", "501", "502", "503", "504", "505", "506", "507", "508",
+			"510", "511",
```
glibsm, line 210: http package has these defined as `http.Status*`

glibsm (commented):
really going deep with the completion support :) i like it

cmd/observe/observe.go
```diff
@@ -177,14 +180,53 @@ more.`,
...
+		return []string{
+			"connect",
+			"delete",
+			"get",
+			"head",
+			"options",
+			"patch",
+			"post",
+			"put",
+			"trace",
```
kAworu, line 227: nit: HTTP methods are usually seen in all-caps (e.g. "POST" rather "post"). Since `filterByHTTPMethods` does a case insensitive check it wouldn't functionally change anything.

kAworu (approved):
Tested and LGTM!

cmd/observe/observe.go
Rolinh: Yes, as a list of `const`. Not sure using them would make the code more readable and they'd have to be turned to string.

glibsm (approved):
A common usecase when doing http codes is querying for things like `5XX`. 

If someone is debugging a specific HTTP code I don't really see the need to autocomplete all of them since they should already know it.

That said this PR doesn't really hurt anything

cmd/observe/observe.go
glibsm: Fair enough.

A helper function to expand `(400, 418)` into a list of strings from 400 to 418, for instance, would definitely help.

PR #443, opened 2020-12-03
cmd: improve command usage message by grouping related flags

Rolinh:
I wanted to create multiple commits but it turns out to be complicated.
Let's summarize what this PR does:

- Add `cmd/common/template` where a `Usage()` can be used by commands to
  generate a custom template based on the provided `pflag.FlagSets`.
  This allows grouping related flags when displaying help/usage of a
  command or subcommand.
- With the help of `Usage()` described above, the custom usage function
  for the `observer` subcommand can be dropped. It was quite hacky and
  easy to get out of sync by adding a flag to the `observe` command and
  forgetting to update the usage formatting function to indicate to
  which flag group the newly added flag belongs. The `observe` command
  now has several defines several `pflag.FlagSet` for each flag group
  (selectors, filters, formatting, others).
- Global flags (ie flags that are persisted from the root command) have
  been split into to `pflag.FlagSet` in `cmd/common/config`:
  `GlobalFlags` and `ServerFlags`. The former contains all global flags
  (ie: debug and config) and the latter all flags related to connecting
  to a Hubble server instance.
- The global flags section is always printed for each command or
  subcommand as it apply to any command. However, the new `ServerFlags`
  flag set needs to be added to the help message by any command or
  subcommand that makes use of it. For the user, this is really
  convenient as these flags appear in the help message of a command only
  when relevant instead of always (eg: the `completion` command does not
  use the server flags and this does not display help for these when
  running `hubble completion -h`). The `ServerFlags` are nonetheless
  global in the sense that they are persistent flags added to the root
  command. This is done for 2 reasons: not breaking the existing
  behavior and allowing users to use these flags in any order (e.g.
  `hubble --server foo observe` also works).
- Instead of using hardcoded strings for every configuration key in
  every command where they need to be used, constants have been defined
  in `cmd/common/config` and shall be used instead.
- Defaults for the configuration directory, fallback configuration
  directory and configuration file have been moved to `pkg/defaults`.

EDIT: sometimes, it's clearer to understand with examples.
This is the help for `observe`, (the grouping of flags is now automatic and based on which flagset the flag is added):
```
$ hubble observe -h
Observe provides visibility into flow information on the network and
application level. Rich filtering enable observing specific flows related to
more.

Usage:
  hubble observe [flags]

Selectors Flags:
      --all            Get all flows stored in Hubble's buffer
  -f, --follow         Follow flows output
      --last uint      Get last N flows stored in Hubble's buffer (default 20)
      --since string   Filter flows since a specific date (relative or RFC3339)
      --until string   Filter flows until a specific date (relative or RFC3339)

Filters Flags:
      --fqdn filter             Show all flows related to the given fully qualified domain name (e.g. "*.cilium.io").
      --from-fqdn filter        Show all flows originating at the given fully qualified domain name (e.g. "*.cilium.io").
      --from-identity filter    Show all flows originating at an endpoint with the given security identity
      --from-ip filter          Show all flows originating at the given IP address.
      --from-label filter       Show only flows originating in an endpoint with the given labels (e.g. "key1=value1", "reserved:world")
      --from-namespace filter   Show all flows originating in the given Kubernetes namespace.
      --from-pod filter         Show all flows originating in the given pod name ([namespace/]<pod-name>). If namespace is not provided, 'default' is used
      --from-port filter        Show only flows with the given source port (e.g. 8080)
      --from-service filter     Show all flows originating in the given service ([namespace/]<svc-name>). If namespace is not provided, 'default' is used
      --http-method filter      Show only flows which match this HTTP method (e.g. "get", "post")
      --http-path filter        Show only flows which match this HTTP path regular expressions (e.g. "/page/\\d+")
      --http-status filter      Show only flows which match this HTTP status code prefix (e.g. "404", "5+")
      --identity filter         Show all flows related to an endpoint with the given security identity
      --ip filter               Show all flows related to the given IP address.
  -l, --label filter            Show only flows related to an endpoint with the given labels (e.g. "key1=value1", "reserved:world")
  -n, --namespace filter        Show all flows related to the given Kubernetes namespace.
      --node-name filter        Show all flows which match the given node names (e.g. "k8s*", "test-cluster/*.company.com")
      --not filter[=true]       Reverses the next filter to be blacklist i.e. --not --from-ip 2.2.2.2
      --pod filter              Show all flows related to the given pod name ([namespace/]<pod-name>). If namespace is not provided, 'default' is used
      --port filter             Show only flows with given port in either source or destination (e.g. 8080)
      --protocol filter         Show only flows which match the given L4/L7 flow protocol (e.g. "udp", "http")
      --service filter          Show all flows related to the given service ([namespace/]<svc-name>). If namespace is not provided, 'default' is used
      --to-fqdn filter          Show all flows terminating at the given fully qualified domain name (e.g. "*.cilium.io").
      --to-identity filter      Show all flows terminating at an endpoint with the given security identity
      --to-ip filter            Show all flows terminating at the given IP address.
      --to-label filter         Show only flows terminating in an endpoint with given labels (e.g. "key1=value1", "reserved:world")
      --to-namespace filter     Show all flows terminating in the given Kubernetes namespace.
      --to-pod filter           Show all flows terminating in the given pod name ([namespace/]<pod-name>). If namespace is not provided, 'default' is used
      --to-port filter          Show only flows with the given destination port (e.g. 8080)
      --to-service filter       Show all flows terminating in the given service ([namespace/]<svc-name>). If namespace is not provided, 'default' is used
  -t, --type filter             Filter by event types TYPE[:SUBTYPE] ([policy-verdict drop debug capture trace l7 agent])
      --verdict filter          Show only flows with this verdict [FORWARDED, DROPPED, ERROR]

Formatting Flags:
      --compact           Deprecated. Use '--output compact' instead.
      --dict              Deprecated. Use '--output dict' instead.
      --ip-translation    Translate IP addresses to logical names such as pod name, FQDN, ... (default true)
  -j, --json              Deprecated. Use '--output json' instead.
      --numeric           Display all information in numeric form
  -o, --output string     Specify the output format, one of:
                           compact:  Compact output
                           dict:     Each flow is shown as KEY:VALUE pair
                           json:     JSON encoding
                           jsonpb:   Output each GetFlowResponse according to proto3's JSON mapping
                           table:    Tab-aligned columns
      --print-node-name   Print node name in output

Server Flags:
      --server string                 Address of a Hubble server (default "unix:///var/run/cilium/hubble.sock")
      --timeout duration              Hubble server dialing timeout (default 5s)
      --tls                           Specify that TLS must be used when establishing a connection to a Hubble server.
                                      By default, TLS is only enabled if the server address starts with 'tls://'.
      --tls-allow-insecure            Allows the client to skip verifying the server's certificate chain and host name.
                                      This option is NOT recommended as, in this mode, TLS is susceptible to machine-in-the-middle attacks.
                                      See also the 'tls-server-name' option which allows setting the server name.
      --tls-ca-cert-files strings     Paths to custom Certificate Authority (CA) certificate files.The files must contain PEM encoded data.
      --tls-client-cert-file string   Path to the public key file for the client certificate to connect to a Hubble server (implies TLS).
                                      The file must contain PEM encoded data.
      --tls-client-key-file string    Path to the private key file for the client certificate to connect a Hubble server (implies TLS).
                                      The file must contain PEM encoded data.
      --tls-server-name string        Specify a server name to verify the hostname on the returned certificate (eg: 'instance.hubble-relay.cilium.io').

Other Flags:
  -s, --silent-errors   Silently ignores errors and warnings

Global Flags:
      --config string   Optional config file (default "/home/isovalent/.config/hubble/config.yaml")
  -D, --debug           Enable debug messages

Get help:
  -h, --help    Help for any command or subcommand
```
This is the help for `config`. Notice that the `Server Flags` section is absent.
```
$ hubble config -h
Config allows to modify or view the hubble configuration. Global hubble options
can be set via flags, environment variables or a configuration file. The
following precedence order is used:

1. Flag
2. Environment variable
3. Configuration file
4. Default value

The "config view" subcommand provides a merged view of the configuration. The
"config set" and "config reset" subcommand modify values in the configuration
file.

Environment variable names start with HUBBLE_ followed by the flag name
capitalized where eventual dashes ('-') are replaced by underscores ('_').
For example, the environment variable that corresponds to the "--server" flag
is HUBBLE_SERVER. The environment variable for "--tls-allow-insecure" is
HUBBLE_TLS_ALLOW_INSECURE and so on.

Usage:
  hubble config [flags]
  hubble config [command]

Available Commands:
  get         Get an individual value in the hubble config file
  reset       Reset all or an individual value in the hubble config file
  set         Set an individual value in the hubble config file
  view        Display merged configuration settings

Global Flags:
      --config string   Optional config file (default "/home/robin/.config/hubble/config.yaml")
  -D, --debug           Enable debug messages

Get help:
  -h, --help    Help for any command or subcommand

Use "hubble config [command] --help" for more information about a command.
```


cmd/common/template/usage_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/cilium/hubble/pkg/defaults"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/stretchr/testify/require"
+)
+
+func TestUsage(t *testing.T) {
+	cmd := &cobra.Command{
+		Use:     "cmd",
+		Aliases: []string{"and", "conquer"},
```
glibsm, line 30: nice!

cmd/node/list.go
```diff
@@ -53,20 +56,24 @@ func newListCommand(vp *viper.Viper) *cobra.Command {
...
 		return []string{
 			"json",
 			"table",
 			"wide",
 		}, cobra.ShellCompDirectiveDefault
 	})
-	return cmd
+	listCmd.Flags().AddFlagSet(formattingFlags)
+
+	listCmd.SetUsageTemplate(template.Usage(formattingFlags, config.ServerFlags))
```
glibsm, line 75: We may want to look into wrapping this in some other function that auto-adds relevant global flags, if that makes sense. I'd rather pass in the `*cmd.Command` into the usage, than pass `template.Usage` into `cmd` if that makes sense.

In that function we can look at the name of the command and take some appropriate steps perhaps?

cmd/common/template/usage_test.go
Rolinh: I was expecting you to catch this one :)

cmd/node/list.go
Rolinh: This is what I wanted to do initially but it turns out that [`AddFlagSet(newSet *FlagSet)`](https://godoc.org/github.com/spf13/pflag#FlagSet.AddFlagSet) flattens the provided `FlagSet`. This means that one can never recover the different flags sets (and name) once they're added to a command (which is rather unfortunate) given the `*cmd.Command` only.

glibsm: I think you may have misunderstood what I was proposing. I wrote up some pseudo code to try to express what I think is possible (excuse the rough syntax please):

```go
// common file

var ServerFlags ...

func Usage() {
	// ...
}

func AddUsage(cmd*cobra.Command, flags ...) {
	var extra []flag

	// some command-specific flags
	if cmd.Name() == "observe" {
		extra = append(extra, ServeFlags)
	}

	extra = append(extra, ConfigFlags)
	cmd.SetUsageTemplate(Usage(flags..., extra))
}

// ------------------------------------

// command file

func NewCommand() {
	cmd := &cobra.Command{...}

	set1 := cmd.Flags()...
	set2 := cmd.Flags()...

	common.AddUsage(cmd, set1, set2)
}
```

I think the main benefit is that `config.ServerFlags` gets auto-added to all the relevant commands in the same place.

Rolinh: I indeed misunderstood. This is a very nice idea, will implement!

Rolinh: @glibsm I gave it a go but it turns out to be problematic for sub-subcommands (e.g. `hubble nodes list`). At creation time, such a sub-subcommand is not attached to a parent command and it's name is simply `list` which is not unique enough to identify. I've tried several variants of the above but in the end, the current implementations is easier to use/understand so I'll leave it at that.

glibsm: Thanks for giving it a shot and proving the update.

gandro (approved):
Awesome! Due the the large size of the PR, it's hard to track what are actual code changes and what just moves code around, but I did try out the branch locally with a bunch of commands and it worked as expected.

glibsm (approved):

PR #444, opened 2020-12-04
completion: remove the copyright header

kAworu:
Before this patch, the zsh completion was broken.

For zsh specifically, the first line of the completion file [must be a zsh autoload tag](http://zsh.sourceforge.net/Doc/Release/Completion-System.html#Autoloaded-files):

>  Files whose first line does not start with one of these tags are not considered to be part of the completion system and will not be treated specially.

### DIFF
--- a/cmd/completion/completion.go
+++ b/cmd/completion/completion.go
@@ -21,22 +21,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const copyRightHeader = `
-# Copyright 2019 Authors of Hubble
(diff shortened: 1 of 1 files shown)


twpayne (approved):

twpayne:
Thanks for finding the solution here :)

Rolinh (approved):
That's a good catch! Let's mark it for backport to v0.7.

gandro (approved):

PR #442, opened 2020-12-02
Add basic support for agent events

tklauser:
Support for agent events was added to the Hubble API in cilium/cilium#14168
    
Allow to print these messages in the Hubble CLI.

Note that this needs cilium/cilium#14242 to work  properly.

Example output (`hubble observe -t agent`):

```
% ./hubble observe -t agent           
TIMESTAMP             NODE                 TYPE                          DETAILS
Dec  2 13:53:15.272   kind-worker          ENDPOINT_CREATED              id: 2172, pod name: coredns-f9fd979d6-p4cz2, namespace: kube-system
Dec  2 13:53:15.298   kind-worker          IPCACHE_UPSERTED              cidr: 10.244.1.155/32, identity: 3, host ip: 172.18.0.2, encrypt key: 0
Dec  2 13:53:15.299   kind-control-plane   IPCACHE_UPSERTED              cidr: 10.244.1.155/32, identity: 3, host ip: 172.18.0.2, encrypt key: 0
Dec  2 13:53:15.633   kind-control-plane   ENDPOINT_CREATED              id: 793
Dec  2 13:53:16.093   kind-worker          ENDPOINT_CREATED              id: 2068
Dec  2 13:53:16.268   kind-worker          ENDPOINT_CREATED              id: 2456, pod name: hubble-relay-5d86f9dc74-np79w, namespace: kube-system
Dec  2 13:53:16.287   kind-control-plane   IPCACHE_UPSERTED              cidr: 10.244.1.105/32, identity: 3, host ip: 172.18.0.2, encrypt key: 0
Dec  2 13:53:16.287   kind-worker          IPCACHE_UPSERTED              cidr: 10.244.1.105/32, identity: 3, host ip: 172.18.0.2, encrypt key: 0
Dec  2 13:53:19.048   kind-control-plane   ENDPOINT_REGENERATE_SUCCESS   id: 2440, labels: [reserved:host,k8s:node-role.kubernetes.io/master]
Dec  2 13:53:19.057   kind-control-plane   ENDPOINT_REGENERATE_SUCCESS   id: 2440, labels: [reserved:host,k8s:node-role.kubernetes.io/master]
Dec  2 13:53:19.648   kind-worker          ENDPOINT_REGENERATE_SUCCESS   id: 464, labels: [reserved:host]
Dec  2 13:53:19.649   kind-worker          ENDPOINT_REGENERATE_SUCCESS   id: 464, labels: [reserved:host]
[...]

PR #452, opened 2020-12-09
Switch protobuf module to google.golang.org/protobuf 

tklauser:
The first commit is a preparatory cleanup removing `pkg/printer.MaybeTime`. Note that this is an exported func, but there are no known external users, so it should be safe to drop it.

The second commit switches from the deprecated `github.com/golang/protobuf` module to `google.golang.org/protobuf`.


pkg/printer/printer.go
```diff
@@ -161,13 +160,12 @@ func (p *Printer) GetHostNames(f v1.Flow) (string, string) {
 	return src, dst
 }
 
-func fmtTimestamp(ts *timestamp.Timestamp) string {
-	t, err := ptypes.Timestamp(ts)
-	if err != nil || t.IsZero() {
+func fmtTimestamp(ts *timestamppb.Timestamp) string {
+	if !ts.IsValid() {
```
gandro, line 164: Very nice find. For other reviewers: `IsValid` checks for nil and invalid timestamps both.

gandro (approved):

kAworu (approved):

tklauser:
FYI, rebased on top of latest master to resolve conflicts due to merge of #442 and #452.

PR #465, opened 2021-01-14
cmd/observer: add support for agent event sub-type filters

tklauser:
The first commit add test coverage for the functionality changed in the successive commits.
The second commit does some small refactoring to avoid duplicating numeric sub-type parsing.
The third commit introduces the actual agent event sub-type filtering. One caveat was that the monitor API (`github.com/cilium/cilium/pkg/monitor/api`) doesn't provide the sub-type strings a form which would be convenient for CLI arguments (i.e. they contain spaces). Thus, `github.com/cilium/cilium/pkg/monitor/api.AgentNotifications` is duplicated in `cmd/observe` in more convenient form. A test is added to ensure the two maps are kept in sync.

See individual commit messages for more details.

### DIFF
--- a/cmd/observe/observe_filter.go
+++ b/cmd/observe/observe_filter.go
@@ -403,19 +403,17 @@ func (of *observeFilter) set(f *filterTracker, name, val string, track bool) err
 		}
 
 		if len(s) > 1 {
-		OUTER:
 			switch t {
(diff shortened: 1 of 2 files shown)


gandro (approved):
Nice. Awesome work with the extended unit tests!

PR #474, opened 2021-02-01
ci: Add dependabot configuration

twpayne:
This adds Dependabot for Hubble, like https://github.com/cilium/cilium/pull/14694.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,20 @@
+version: 2
+updates:
+  - package-ecosystem: gomod
+    directory: /
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 1
+    rebase-strategy: disabled
+    labels:
+    - kind/enhancement
+    - release-note/misc
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 1
+    rebase-strategy: disabled
+    labels:
+    - kind/enhancement
+    - release-note/misc


.github/dependabot.yml
```diff
@@ -0,0 +1,21 @@
+version: 2
+updates:
+  - package-ecosystem: gomod
+    directory: /
+    schedule:
+      interval: weekly
+    ignore:
+        # cannot be updated until the etcd library is updated
+      - dependency-name: "google.golang.org/grpc"
```
kaworu, line 9: Hubble doesn't have the etcd dependency and the vendored grpc module is up-to-date.
```suggestion
```

kaworu (changes requested):
Thanks for this PR @twpayne!

.github/dependabot.yml
```diff
@@ -0,0 +1,21 @@
...
+        # using a cilium-specific fork
+      - dependency-name: "github.com/miekg/dns"
```
tklauser, line 11: I think this isn't necessary either. `github.com/miekg/dns` is not a direct Hubble dependency.

tklauser (changes requested):

.github/dependabot.yml
twpayne: I left this in because the `github.com/miekg/dns` override is explicitly in `go.mod`, but I'll try removing it.

.github/dependabot.yml
twpayne: OK, I've removed this.

tklauser (approved):
LGTM! Should we disable automatic rebasing as well (see https://github.com/cilium/cilium/pull/14826)? Though, in Hubble we don't have long running Jenkins CI jobs which would be affected, but it might lead to fewer surprises compared to dependabot behaviour in `cilium/cilium`.

gandro (approved):
LGTM as well! I agree with Tobias that we probably don't need automatic rebase.

tklauser:
FYI: https://github.com/cilium/cilium/pull/14837 will probably make sense to limit the number of open PRs for the Hubble repo as well to avoid merge conflicts between update PRs.

twpayne:
OK, PR update to disable rebases and limit the number of open PRs to 1, as proposed by @tklauser.

kaworu:
Does this mean that dependabot can only open one PR per week? If correct, is there a scenario where it can't keep up with the dependencies update (i.e. there are more than one module update per week)?

tklauser:
> Does this mean that dependabot can only open one PR per week? If correct, is there a scenario where it can't keep up with the dependencies update (i.e. there are more than one module update per week)?

As I read the [dependabot config documentation](https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates), it will check for updates once a week and then send them one by one. So there could be multiple update PRs in a week but only one open at a time. They will be updating the individual modules to the most recent version at the time of the update check, which could in the worst case be one week old.

kaworu (approved):
@tklauser thanks for looking into this, additionaly the bits about security fixes (which was my main concern) having their separate limit is nice:
> [_about open-pull-requests-limit_] This option has no impact on security updates, which have a separate, internal limit of ten open pull requests.

PR #475, opened 2021-02-01
ci: Add CodeQL analysis

twpayne:
This adds CodeQL analysis to Hubble, similar to https://github.com/cilium/cilium/pull/14514.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,32 @@
+name: codeql
+
+on:
+  push:
+    branches:
+    - master
+    - v0.7
+    - v0.6
+    - v0.5
+  pull_request:
+    branches:
+    - master
+  schedule:
+    - cron: "45 7 * * 3"
+
+jobs:
+  analyze:
+    if: github.repository == 'cilium/hubble'
+    runs-on: ubuntu-18.04
+    strategy:
+      fail-fast: false
+    steps:
+    - name: Checkout repo
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 1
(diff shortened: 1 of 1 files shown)


kaworu:
Thanks @twpayne for the PR! Is the "analysis not found" expected? https://github.com/cilium/hubble/pull/475/checks?check_run_id=1807367595

twpayne:
Nope, definitely not expected. With these CI changes, you only get to debug them once you push a branch :)

kaworu:
So what would be the way to make forward progress on this?

twpayne:
> So what would be the way to make forward progress on this?

I need to do some debugging :)

twpayne:
I think this is just because this is the first CodeQL analysis for this repo, so there's no other analysis to compare with to find what's been fixed or newly introduced. [Relevant doc](https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/enabling-code-scanning-for-a-repository#reasons-for-the-missing-analysis-message).

kaworu:
> I think this is just because this is the first CodeQL analysis for this repo, so there's no other analysis to compare with to find what's been fixed or newly introduced. Relevant doc.

That make sense. How about this: we merge this PR and you make another (dummy or cosmetic) PR to see if everything work as expected?

kaworu (approved):

gandro (approved):

twpayne:
> That make sense. What we could do is merge this PR, and then you make another (dummy or cosmetic) PR to see if everything work as expected. How does that sound?

Good idea. I've rebased #474 on top of this, which should serve the purpose.

PR #477, opened 2021-02-05
ci: fix dependabot kind/enhancement label

kaworu:
Before this patch, because the kind/enhancement label is prefixed by the `chart_with_upwards_trend` icon, dependabot would complain to be unable to find the "kind/enhancement" label. See for example https://github.com/cilium/hubble/pull/476#issuecomment-773892140

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,5 +16,5 @@ updates:
     open-pull-requests-limit: 1
     rebase-strategy: disabled
     labels:
-    - kind/enhancement
-    - release-note/misc
+    - "📈 kind/enhancement"
+    - "release-note/misc"


rolinh (approved):

twpayne (approved):
Thanks, not sure how that snuck in.

twpayne (changes requested):
Hang on, this PR *adds* the emoji. Are you sure that this is correct?

twpayne (approved):
Sorry, I see that we're using emojis in our labels 😱 . Not sure that this is a good idea at all, but until that's resolved I see the need for this fix.

kaworu:
@twpayne if the bot continue to fail to find the label after this PR is merged, I think it will then be reasonable to request for their removal.

PR #485, opened 2021-02-10
ci: do not upload artifacts

rolinh:
See commits for details.


.github/workflows/tests.yml
```diff
@@ -24,14 +24,10 @@ jobs:
       run: make check
     - name: Run unit tests
       run: make test
-    - name: Generate artifacts
+    - name: Build
+      run: make
+    - name: Build release binaries
```
gandro, line 29: Out of scope for this PR, but I think the original intent here was automating the binary generation. We should see how hard it would be to automatically create a release via GH API for each newly created `v*` tag.

gandro (approved):

PR #490, opened 2021-02-17
Automate release creation and artifacts publishing

rolinh:
See commits for details. As release notes need to be manually edited, new releases are created as draft so that a maintainer can edit the release notes before publishing.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,38 @@
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:
+      - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
+
+name: Create a release
+
+jobs:
+  build:
+    name: Create Release
+    if: github.repository == 'cilium/hubble'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Generate artifacts
+        run: make release
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # This token is provided by Actions, you do not need to create your own token
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: ${{ github.ref }}
+          draft: true # turn this to false once release notes are automatically added
+          prerelease: false
+          body: |
+            Note for maintainers:: Please, update the desciption with the actual release notes (see RELEASE.md for instructions).
+      - name: Upload artifacts 
+        id: upload-release-artifacts
+        uses: skx/github-action-publish-binaries@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          releaseId: ${{ steps.create_release.outputs.id }}
+          args: 'release/*'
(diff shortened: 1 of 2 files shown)


RELEASE.md
```diff
@@ -103,31 +103,13 @@ The README file contains a section which lists all currently supported releases
...
 release.
 
-## Create a GitHub release
+## Update the GitHub release notes
 
-It is better to have github create the final release tag, rather than pushing
-it through git. Pushing through git will auto-create an empty release and
-notify all the users before you have a chance to include the list of changes,
-or any other metadata.
+When a tag is pushed, a Github Action job takes care of creating a new Github
```
gandro, line 108: Should we add the instructions to push the tag via git?

.github/workflows/release.yml
```diff
@@ -0,0 +1,39 @@
...
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # This token is provided by Actions, you do not need to create your own token
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: ${{ github.ref }}
+          draft: true # turn this to false once release notes are automatically added
+          prerelease: false
+          body: |
+            Note for maintainers:: Please, update the desciption with the actual release notes (see RELEASE.md for instructions).
```
gandro, line 31: Is the double colon intentional here?

.github/workflows/release.yml
```diff
@@ -0,0 +1,39 @@
...
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:
+      - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
+
+name: Create a release
+
+jobs:
+  build:
+    name: Create Release
```
gandro, line 11: Do we want to add an `if: ${{ github.repository == 'cilium/hubble' }}` here to avoid this job being executed on forks?

RELEASE.md
rolinh: I think there are instructions above in this file.

.github/build
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+make release
```
rolinh, line 2: note: I actually think this will fail because it invokes docker.... but I'm not sure. I'd be ok to invoke `make local-release` as long as we can guarantee the Go version that is used to build the release.

.github/workflows/release.yml
rolinh: Yes, just not get my YAML syntax highlighter confused :innocent:

RELEASE.md
rolinh: I was wrong. I added a new step with instructions to create and push a tag.

.github/build
gandro: Looking at the source code, I'm not sure what exactly the build action does, it just seems to invoke this script, nothing else: https://github.com/skx/github-action-build

Maybe we can run the build with a normal shell step into the `release/` folder? 

For sure, we probably want to test this on a throw-away repo first before merging it, to avoid spamming subscribed users with test tags.

rolinh: OK, I tested this on a fork. The artifacts generation step indeed failed but it turns out that `skx/github-action-build` does nothing special and using a simple `make release` as a step instead is enough and works just fine. I think this PR is good to go as is.

gandro (approved):
Nice work!

kaworu (approved):
Great work, thanks!

PR #499, opened 2021-03-02
cmd/peer: Refactor and test processing of response

simar7:
This PR refactors a small piece of processing code and adds tests.

Signed-off-by: Simarpreet Singh <simar@linux.com>

### DIFF
--- a/cmd/peer/watch.go
+++ b/cmd/peer/watch.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"os"
 
 	peerpb "github.com/cilium/cilium/api/v1/peer"
 	"github.com/cilium/hubble/cmd/common/config"
@@ -58,11 +59,7 @@ func runWatch(ctx context.Context, client peerpb.PeerClient) error {
 		case io.EOF, context.Canceled:
 			return nil
 		case nil:
-			tlsServerName := ""
-			if tls := resp.GetTls(); tls != nil {
-				tlsServerName = fmt.Sprintf(" (TLS.ServerName: %s)", tls.GetServerName())
-			}
-			fmt.Printf("%-12s %s %s%s\n", resp.GetType(), resp.GetAddress(), resp.GetName(), tlsServerName)
+			processResponse(os.Stdout, resp)
 		default:
 			if status.Code(err) == codes.Canceled {
 				return nil
@@ -70,4 +67,12 @@ func runWatch(ctx context.Context, client peerpb.PeerClient) error {
 			return err
 		}
 	}
+}
+
+func processResponse(w io.Writer, resp *peerpb.ChangeNotification) {
+	tlsServerName := ""
+	if tls := resp.GetTls(); tls != nil {
+		tlsServerName = fmt.Sprintf(" (TLS.ServerName: %s)", tls.GetServerName())
+	}
+	_, _ = fmt.Fprintf(w, "%-12s %s %s%s\n", resp.GetType(), resp.GetAddress(), resp.GetName(), tlsServerName)
 }
--- a/cmd/peer/watch_test.go
+++ b/cmd/peer/watch_test.go
@@ -1,0 +1,61 @@
+// Copyright 2021 Authors of Hubble
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package peer
+
+import (
+	"bytes"
+	"testing"
+
+	peerpb "github.com/cilium/cilium/api/v1/peer"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_processResponse(t *testing.T) {
+	var testCases = []struct {
+		name               string
+		changeNotification *peerpb.ChangeNotification
+		expectedOutput     string
+	}{
+		{
+			name: "happy path with tls",
+			changeNotification: &peerpb.ChangeNotification{
+				Name:    "foo.bar",
+				Address: "1.2.3.4",
+				Type:    peerpb.ChangeNotificationType_PEER_ADDED,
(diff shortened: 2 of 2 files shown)


simar7:
Some insight: Initially I wanted to treat the `runWatch` function as a black box and write a test for it to simulate state changes. While it was exciting at first, the test setup for this proved to be quite involved since `runWatch` runs a busy loop and is only controllable through the error return value from `b.Recv()`. 

In favour of simplicity and readability, I instead took this approach. This refactors and tests the functional aspects of processing an event, rather than its setup.

rolinh:
> Initially I wanted to treat the runWatch function as a black box and write a test for it to simulate state changes. While it was exciting at first, the test setup for this proved to be quite involved since runWatch runs a busy loop and is only controllable through the error return value from b.Recv().

While not trivial, it's probably simpler than you think. We actually have[ a fake peer client in the Cilium tree](https://github.com/cilium/cilium/blob/198fe64d79b53c983ba74f09f5c7a1c34408666d/pkg/hubble/testutils/fake.go#L132-L159). It is actually used for [Hubble Relay's pool manager unit tests](https://github.com/cilium/cilium/blob/master/pkg/hubble/relay/pool/manager_test.go).
As part of the test, you can simply instantiate a fake peer client add pass it to `runWatch()`. You don't have to go this route though but it should actually be fairly simple to implement as all that needs to be done is return something when the fake peer client's `Notify` method is called.

simar7:
> > Initially I wanted to treat the runWatch function as a black box and write a test for it to simulate state changes. While it was exciting at first, the test setup for this proved to be quite involved since runWatch runs a busy loop and is only controllable through the error return value from b.Recv().
> 
> While not trivial, it's probably simpler than you think. We actually have[ a fake peer client in the Cilium tree](https://github.com/cilium/cilium/blob/198fe64d79b53c983ba74f09f5c7a1c34408666d/pkg/hubble/testutils/fake.go#L132-L159). It is actually used for [Hubble Relay's pool manager unit tests](https://github.com/cilium/cilium/blob/master/pkg/hubble/relay/pool/manager_test.go).
> As part of the test, you can simply instantiate a fake peer client add pass it to `runWatch()`. You don't have to go this route though :) But it should actually be fairly simple to implement as all that needs to be done is to return something when the fake peer client's `Notify` method is called.

Thanks for that insight @rolinh – I unknowingly did venture into that route on my own before charting on the current way but I will re-visit it once again. The pool manager unit tests look very interesting in their approach. Definitely something to learn from.

simar7:
So I tried to use the testutils package as is it has the test doubles needed in this case but I ran into a small hiccup. I believe the package isn't importable outside the repository as it doesn't have a `go.mod` within the package. https://github.com/cilium/cilium/tree/master/pkg/hubble/testutils

I tried to validate the above by doing the following, which seems to work:

1. Locally cloning a copy of cilium repo
2. Adding the appropriate `go.mod` 
```
$ testutils git:(master) ✗ cat go.mod 
module github.com/pkg/hubble/testutils

go 1.16
```
3. Doing a `replace` directive to point at the locally cloned cilium repo. 
```
replace (
	github.com/cilium/cilium/pkg/hubble/testutils => /Users/simar/repos/cilium/pkg/hubble/testutils
)
```

I believe testutils may or may not have been intended to use outside of the cilium repo and as a result it isn't importable as a go module. It should be a simple addition to make `testutils` importable and I can contribute a PR for it if that's OK.

In the interest of slicing thin, do you think we should keep adding tests for `runWatch()` in a separate PR and merge this if it looks good to you as is?

rolinh:
> I believe the package isn't importable outside the repository as it doesn't have a go.mod within the package.

Actually, this is not the issue at play here. We suffer from a problem with the grpc dependency which broke version compatibility and this affects a cilium dependency (etcd). We managed to use a recent grpc version in this repository because so far, nothing we pulled from cilium pulled the etcd dependency. It seems that pulling in the `testutils` package does :/ :
```
$ go mod tidy && go mod vendor && go mod verify
go: finding module for package google.golang.org/grpc/naming
github.com/cilium/hubble/cmd/peer tested by
        github.com/cilium/hubble/cmd/peer.test imports
        github.com/cilium/cilium/pkg/hubble/testutils imports
        github.com/cilium/cilium/pkg/ipcache imports
        github.com/cilium/cilium/pkg/kvstore imports
        go.etcd.io/etcd/clientv3 tested by
        go.etcd.io/etcd/clientv3.test imports
        go.etcd.io/etcd/integration imports
        go.etcd.io/etcd/proxy/grpcproxy imports
        google.golang.org/grpc/naming: module google.golang.org/grpc@latest found (v1.36.0), but does not contain package google.golang.org/grpc/naming
``` 
More information about this problem [here](https://github.com/cilium/cilium/pull/13405#issuecomment-704766707). We're currently pending on a new release of the etcd dependency which should fix this problem.

> In the interest of slicing thin, do you think we should keep adding tests for runWatch() in a separate PR and merge this if it looks good to you as is?

As I can't tell when the above issue will be resolved, I think your proposal is reasonable.

rolinh (approved):

simar7:
> > I believe the package isn't importable outside the repository as it doesn't have a go.mod within the package.
> 
> Actually, this is not the issue at play here. We suffer from a problem with the grpc dependency which broke version compatibility and this affects a cilium dependency (etcd). We managed to use a recent grpc version in this repository because so far, nothing we pulled from cilium pulled the etcd dependency. It seems that pulling in the `testutils` package does :/ :
> 
> ```
> $ go mod tidy && go mod vendor && go mod verify
> go: finding module for package google.golang.org/grpc/naming
> github.com/cilium/hubble/cmd/peer tested by
>         github.com/cilium/hubble/cmd/peer.test imports
>         github.com/cilium/cilium/pkg/hubble/testutils imports
>         github.com/cilium/cilium/pkg/ipcache imports
>         github.com/cilium/cilium/pkg/kvstore imports
>         go.etcd.io/etcd/clientv3 tested by
>         go.etcd.io/etcd/clientv3.test imports
>         go.etcd.io/etcd/integration imports
>         go.etcd.io/etcd/proxy/grpcproxy imports
>         google.golang.org/grpc/naming: module google.golang.org/grpc@latest found (v1.36.0), but does not contain package google.golang.org/grpc/naming
> ```
> 
> More information about this problem [here](https://github.com/cilium/cilium/pull/13405#issuecomment-704766707). We're currently pending on a new release of the etcd dependency which should fix this problem.
> 
> > In the interest of slicing thin, do you think we should keep adding tests for runWatch() in a separate PR and merge this if it looks good to you as is?
> 
> As I can't tell when the above issue will be resolved, I think your proposal is reasonable.

Oh I see. I'll subscribe to that PR in that case to keep track.

PR #502, opened 2021-03-04
cmd/observe: print filters in debug mode

rolinh:
When debugging, it can be really useful to know about the filters that have been generated.

Example:

    $ hubble --debug observe --ip 9.9.9.9  > /dev/null
    Using config file: /home/user/.config/hubble/config.yaml
    Using filters:
    => include: [{"source_ip":["9.9.9.9"]},{"destination_ip":["9.9.9.9"]}]
    => exclude: <nil>

### DIFF
--- a/cmd/observe/observe.go
+++ b/cmd/observe/observe.go
@@ -95,8 +95,12 @@ application level. Rich filtering enable observing specific flows related to
 individual pods, services, TCP connections, DNS queries, HTTP requests and
 more.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if err := handleArgs(ofilter, vp.GetBool(config.KeyDebug)); err != nil {
+			debug := vp.GetBool(config.KeyDebug)
+			if err := handleArgs(ofilter, debug); err != nil {
 				return err
+			}
+			if debug {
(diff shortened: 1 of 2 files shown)


michi-covalent (approved):

cmd/observe/observe.go
```diff
@@ -95,9 +95,13 @@ application level. Rich filtering enable observing specific flows related to
 individual pods, services, TCP connections, DNS queries, HTTP requests and
 more.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if err := handleArgs(ofilter, vp.GetBool(config.KeyDebug)); err != nil {
+			debug := vp.GetBool(config.KeyDebug)
+			if err := handleArgs(ofilter, debug); err != nil {
 				return err
 			}
+			if debug {
+				fmt.Fprintf(cmd.ErrOrStderr(), "Using filters:\n=> whitelist: %s\n=> blacklist: %s\n", ofilter.whitelist, ofilter.blacklist)
```
gandro, line 103:
```suggestion
				fmt.Fprintf(cmd.ErrOrStderr(), "Using filters:\n=> include: %s\n=> exclude: %s\n", ofilter.whitelist, ofilter.blacklist)
```

nit: While the API field itself is called `blacklist/whitelist` and therefore cannot easily be renamed, we have been trying to move away from the blacklist/whitelist terminology.

gandro (approved):

PR #506, opened 2021-03-09
cmd/observe: mark deprecated output flags as deprecated

rolinh:
This hides the flags "json", "compact" and "dict" from the help output and prints the replacement flag to use instead. E.g. for "json":

    $ hubble observe --json
    Flag --json has been deprecated, use '--output json' instead


glibsm (approved):

glibsm:
I think this is not a `misc` note, since this is something user-facing and should appear as a minor release note.

PR #516, opened 2021-03-20
Move version into VERSION file

glibsm:
Looks like the v0.7 version was wrong, so I corrected that as well. `0.7` branch shouldn't report version as `0.7.1`, as that should only appear in the release commit.


rolinh (approved):

kaworu (approved):

gandro (changes requested):
I'm a bit torn about this change. The `0.7.1` is intentional, we're only been using `-dev` for master so far. Which is consistent with how Cilium does it, e.g. compare:

https://github.com/cilium/cilium/blob/v1.8/VERSION
https://github.com/cilium/cilium/blob/master/VERSION

I'm fine with changing the way we manage the VERSION file, but this needs to be reflected in the `RELEASE.md` process, as I'm not sure how I can do a point release with the new approach (i.e. do I need to revert back to 0.7-dev after every release?)

glibsm:
> The 0.7.1 is intentional

The `0.7.1` can not be intentional. A branch `vX.Y` tagged with version `vX.Y.Z` matching a previous release `vX.Y.Z` loses all meaning because there can be any number of commits in between.

The `.90` I'm familiar with but so far nobody has been able to explain the rational for it very well.

To me `-dev` postfix on branches in between releases makes the most sense. If it's really that big a deal we can use `.90` instead of `-dev` but it can not be left a `0.7.1`

michi-covalent:
`-dev` makes sense to me, it would be good to update `RELEASE.md`.

the pattern i've seen in the past goes something like this:

- version gets set to something like `0.7.1-dev`. it means it's a dev version of `0.7.1` which hasn't been released yet.
- when you make a release, set the version to `0.7.1`.
- after release is done, set the version to `0.7.2-dev`. now it's a dev version for the next release `0.7.2`.

gandro:
> > The 0.7.1 is intentional
> 
> The `0.7.1` can not be intentional. A branch `vX.Y` tagged with version `vX.Y.Z` matching a previous release `vX.Y.Z` loses all meaning because there can be any number of commits in between.

It's intentional in the sense that we are we (or at least I) was consciously mirroring what Cilium is doing. I'm fine with moving away from what Cilium is doing and have the `-dev` suffix in release branches, in that case however we need to update the `RELEASE.md` as to how to update the `VERSION` file for a release as @michi-covalent suggested.

glibsm:
> It's intentional in the sense that we were (or at least I was) consciously mirroring what Cilium is doing

I was just trying to point out that the branch had an error no matter what process we used. The version was still the same from the previous release 🤷‍♂️

image

gandro:
> > It's intentional in the sense that we were (or at least I was) consciously mirroring what Cilium is doing
> 
> I was just trying to point out that the branch had an error no matter what process we used. The version was still the same from the previous release man_shrugging
> 
> image

According to Cilium's process, the `v0.7.1` is correct ;) Please check Cilium's 1.8 branch, you will see that there were also commits since the file was modified for the last release, exactly in the same manner we did here.

As this seems confusing, I'm happy to move away from that process. But I'd like to have documented how one is to update the version file, as now each release needs two commits to the version file.

rolinh (approved):

glibsm:
> According to Cilium's process, the v0.7.1 is correct ;) 

🙀

michi-covalent (approved):

gandro (approved):

PR #517, opened 2021-03-20
vendor: Fix grpc version (1.33 doesn't exist)

glibsm:
Branch currently fails with the following:
```
❯ go mod tidy
go: google.golang.org/grpc@v1.33.0: reading google.golang.org/grpc/go.mod at revision v1.33.0: unknown revision v1.33.0
```

I was trying to find the closest version tat wouldn't
require any other code changes and this seems to be it.

Master is running a 1.29.1 which is also something we could
consider for this branch.

Signed-off-by: Glib Smaga <code@gsmaga.com>


rolinh:
It seems that release 1.33.0 was [published by mistake](https://github.com/grpc/grpc-go/issues/3945) but was picked up. What about bumping to release 1.33.3?

> Master is running a 1.29.1 which is also something we could consider for this branch.

Master is actually running 1.36.0.

kaworu:
> Master is running a 1.29.1 which is also something we could consider for this branch.

I think you are refering to Cilium master. I would agree with @rolinh to consider v1.33.3 instead as going back to v1.29.1 imply loosing descriptive error messages on TLS handshake failures (see https://github.com/cilium/hubble/pull/372 for context).

gandro:
+1 on using 1.33.3 on v0.7 for the better mTLS error messages (which helps a ton when troubleshooting). Assuming 1.33.x doesn't break any other builds of course.

glibsm:
Yes I didn't want to go back to `1.29` as that would require code change and a downgrade in experience. Let me try `1.33.3` to see if it's compatible and an easy fix for a while

gandro (approved):

rolinh (approved):

PR #524, opened 2021-04-02
Support reading flows from stdin

### LINKED ISSUE #523: hubble observe: support reading from json file
say i have the output of `hubble observe -o jsonpb` saved to a file. it would be handy if i can pass the file to `hubble observe` later and somehow apply filters instead of trying to come up with equivalent `jq` queries.

michi-covalent:
Implement an ObserverClient that reads jsonpb flows from an io.Reader
instead of connecting to Hubble, and modify `hubble observe` command
to use it when stdin is a pipe. For example,

    hubble observe -o jsonpb --last 1000 > flows.log
    cat flows.log | hubble observe -n kube-system

A few things to note for observing flows from stdin:

- --follow, --last, and server flags are ignored.
- Flows are processed in the order they are read from stdin without
  sorting them by timestamp.

Fixes #523


cmd/observe/io_reader_observer.go
```diff
@@ -0,0 +1,103 @@
...
+import (
+	"bufio"
+	"context"
+	"io"
+
+	"github.com/cilium/cilium/api/v1/observer"
+	v1 "github.com/cilium/cilium/pkg/hubble/api/v1"
+	"github.com/cilium/cilium/pkg/hubble/filters"
+	"github.com/cilium/hubble/pkg/logger"
+	"github.com/golang/protobuf/jsonpb"
```
rolinh, line 26: Package `github.com/golang/protobuf` is superseded by `google.golang.org/protobuf`. Please, use `google.golang.org/protobuf/encoding/protojson` to avoid importing `github.com/golang/protobuf` altogether.

cmd/observe/io_reader_observer.go
```diff
@@ -0,0 +1,103 @@
...
+	panic("GetNodes is not supported")
+}
+
+func (o *ioReaderObserver) ServerStatus(_ context.Context, _ *observer.ServerStatusRequest, _ ...grpc.CallOption) (*observer.ServerStatusResponse, error) {
+	panic("ServerStatus is not supported")
+}
+
+// ioReaderClient implements Observer_GetFlowsClient.
+type ioReaderClient struct {
+	reader  *bufio.Reader
```
rolinh, line 56: What about using `bufio.Scanner` instead?

rolinh (changes requested):
This is pretty neat, I like it! :rocket:  Please, see some suggestions for improvements below.

Have you considered reading from a (several) file(s) as well? Maybe via optional positional arguments to the `observe` command.

cmd/observe/io_reader_observer.go
```diff
@@ -0,0 +1,103 @@
...
+		if err != nil {
+			logger.Logger.WithError(err).WithField("line", line).Warn("Failed to unmarshal json to flow")
+			continue
+		}
+		if c.request.Since != nil && c.request.Since.AsTime().After(res.Time.AsTime()) {
+			continue
+		}
+		if c.request.Until != nil && c.request.Until.AsTime().Before(res.Time.AsTime()) {
+			continue
+		}
```
gandro, line 97: This doesn't handle `c.request.Number`. Is that intentional?

gandro (commented):
Nice work! :rocket:  Just one question in addition to what Robin already pointed out.

cmd/observe/io_reader_observer.go
michi-covalent: yes it is somewhat intentional. you can always do this instead 😉 
```
hubble observe -n kube-system | tail -n 20
```

michi-covalent:
> Have you considered reading from a (several) file(s) as well? Maybe via optional positional arguments to the `observe` command.

reading from multiple files are already supported in a way. you can do:
```
cat *.log | hubble observe
```
you can also pre-process files before sending to `hubble observe`. for example if flow.log files are compressed:
```
zcat *.log | hubble observe
```
we could add positional arguments, but i thought it might get a bit more confusing to users. right now the separation is pretty clear i think: if stdin is a pipe, read flows from there. else connect to hubble server.

cmd/observe/io_reader_observer.go
```diff
@@ -0,0 +1,103 @@
...
+		if !filters.Apply(c.allow, c.deny, &v1.Event{Timestamp: res.Time, Event: res.GetFlow()}) {
+			continue
+		}
+		return &res, nil
+	}
+	return nil, io.EOF
```
rolinh, line 102: Please, check for errors:
```suggestion
      if err := scanner.Err(); err != nil {
          return nil, err
      }
	return nil, io.EOF
```

rolinh (changes requested):
Lgtm besides a missing error check.

cmd/observe/io_reader_observer.go
michi-covalent: fixed!

rolinh (approved):

cmd/observe/io_reader_observer.go
rolinh: How would this work for our Windows users though? 😀

michi-covalent: that's a good question 😱 not tested, but with powershell something like this might work?
```
gc flow.log | hubble observe -n kube-system | select -last 20
```

PR #529, opened 2021-04-06
Add a hidden --flow-export flag

michi-covalent:
Use this flag to write flows without applying any filters to a rotated
file. The syntax is "$filepath;$max-size-mb;$max-backups". For example:

        --flow-export=/var/run/cilium/flow/flow.log;10;5

writes flows to /var/run/cilium/flow/flow.log, rotate the file when it
reaches 10MB uncompressed, and keep the last 5 rotated files compressed.


pkg/server/unfiltered_exporter.go
```diff
@@ -0,0 +1,113 @@
+// Copyright 2020 Authors of Cilium
```
kaworu, line 1: nit:
```suggestion
// Copyright 2021 Authors of Cilium
```

kaworu (approved):

PR #530, opened 2021-04-13
cmd: Add record subcommand

gandro:
CLI for the Recorder API implemented in https://github.com/cilium/cilium/pull/15680

```console
$ hubble record "0.0.0.0/0 0:0.0.0.0/0 0 ANY"
Started recording. Press CTRL+C to stop.
2021-04-13T20:24:13Z Status: 0 packets (0 bytes) written
2021-04-13T20:24:14Z Status: 3 packets (237 bytes) written
2021-04-13T20:24:16Z Status: 11 packets (996 bytes) written
2021-04-13T20:24:18Z Status: 61 packets (6775 bytes) written
2021-04-13T20:24:19Z Status: 66 packets (7610 bytes) written
2021-04-13T20:24:21Z Status: 70 packets (8090 bytes) written
^CStopping recording...
2021-04-13T20:24:25Z Status: 70 packets (8090 bytes) written
2021-04-13T20:24:25Z Output: /var/run/cilium/pcaps/hubble_1618345449_4039455774_k8s1.pcap
```


cmd/record/record.go
```diff
@@ -0,0 +1,217 @@
...
+// New creates a new record subcommand
+func New(vp *viper.Viper) *cobra.Command {
+	recordCmd := &cobra.Command{
+		Use:   "record [flags] filter1 filter2 ... filterN",
+		Short: "Capture and record network packets",
+		Long:  "Capture and record network packets into a pcap file stored on the Hubble server",
+		Example: `
+Record all TCP traffic from 192.168.1.0/24 (any source port) to 10.0.0.0/16 port 80:
+
+  hubble record "192.168.1.0/24 0 10.0.0.0/16 80 TCP"
```
glibsm, line 55: commit contains a different example `hubble record "0.0.0.0/0:0:0.0.0.0/0:0:ANY"`

cmd/record/record.go
```diff
@@ -0,0 +1,217 @@
...
+		dstPort, err := strconv.ParseUint(f[3], 10, 16)
+		if err != nil {
+			return nil, fmt.Errorf("invalid destination port: %s", f[3])
+		}
+		proto, err := parseProto(f[4])
+		if err != nil {
+			return nil, err
+		}
+
+		result = append(result, &recorderpb.Filter{
```
glibsm, line 130: is the plan to leave all the error handling for the server, or add some validation in the client side too?

cmd/record/record.go
```diff
@@ -0,0 +1,217 @@
...
+	recorderFlags.StringVar(&fileSinkPrefix, "file-prefix", "hubble", "File prefix of the resulting pcap file")
+	recorderFlags.Uint32Var(&maxCaptureLength, "max-capture-len", 0, "Sets the maximum capture length (zero for full capture)")
+
+	recordCmd.Flags().AddFlagSet(recorderFlags)
+	recordCmd.SetUsageTemplate(template.Usage(config.ServerFlags, recorderFlags))
+
+	return recordCmd
+}
+
+func parseProto(proto string) (recorderpb.Protocol, error) {
```
glibsm, line 90: does proto not generate a function similar to this already?

cmd/record/record.go
```diff
@@ -0,0 +1,217 @@
...
+
+	statusPrinted := false
+	for {
+		stats, err := recorder.Recv()
+		switch {
+		case err != nil:
+			return err
+		case stats.GetRunning() != nil:
+			if isTTY && statusPrinted {
+				fmt.Print(resetLastLine)
```
glibsm, line 206: not clear to me what this does, exactly

cmd/record/record.go
gandro: We also don't so client-side validation for almost all filters in the normal observe API, so I think it's fine to leave this server-side, also to avoid the two implementations diverging.

cmd/record/record.go
gandro: This clears/resets the current line in stdout if stdout is a terminal similar to how tcpdump works. Probably best explained via video:

https://user-images.githubusercontent.com/50564/115267955-0fc9f180-a13a-11eb-9260-1255623277a6.mp4

cmd/record/record.go
gandro: Not to my knowledge.

cmd/record/record.go
gandro: Updated the PR. I moved away from `:` due to ambiguity with IPv6 addresses. We could use the `[ffff::1]:0` notation, but it's more annoying to type.

cmd/record/record.go
```diff
@@ -0,0 +1,217 @@
...
+			output, stats.GetPacketsLost(), stats.GetBytesLost())
+	}
+	return output
+}
+
+func runRecord(ctx context.Context, conn *grpc.ClientConn, filters []*recorderpb.Filter) error {
+	isTTY := false
+	if fileInfo, _ := os.Stdout.Stat(); (fileInfo.Mode() & os.ModeCharDevice) != 0 {
+		isTTY = true
+	}
```
kaworu, line 160: cosmetic suggestion:
```suggestion
	fileInfo, _ := os.Stdout.Stat()
	isTTY := (fileInfo.Mode() & os.ModeCharDevice) != 0
```

Also question: is this implentation portable? Would something like https://pkg.go.dev/golang.org/x/term#IsTerminal make sense since we have OSX and Windows builds?

kaworu (approved):
Great work @gandro! Only have a couple of questions, feel free to ignore.

cmd/record/record.go
gandro: Hm. So we want to make sure that right now, `isTTY` is false on Windows, because it will not support the escape code we're using. Unfortunately `golang.org/x/term` does not have escape codes for the line reset (only color), so we probably would have to find a full-fledged library.

I'll do some digging, but I might end up only supporting Linux for now, while other OSes should just have `isTTY=false`, which will always work.

rolinh (commented):
This looks good to me in general. However, I have an API design question: the recorder API accepts a stream of record requests (`rpc Record (stream RecordRequest) returns (stream RecordResponse);`). With the current implementation, only one request can be sent, after which the CLI waits for `ctrl+c`. In this context, how would a user record several sessions? With the current design, it seems that it will be hard to implement such a feature in the future without a breaking change.

gandro:
> This looks good to me in general. However, I have an API design question: the recorder API accepts a stream of record requests (`rpc Record (stream RecordRequest) returns (stream RecordResponse);`). With the current implementation, only one request can be sent, after which the CLI waits for `ctrl+c`. In this context, how would a user record several sessions? With the current design, it seems that it will be hard to implement such a feature in the future without a breaking change.

There are two parts to this questions:

From the users perspective, if a user wants to run two concurrent recordings, the UX will be the same as it is e.g. for `tcpdump`: Start two `hubble record` commands in parallel. This has already been demonstrated to work with the current implementation.

From an API perspective, it is currently an error to submit a `StartRecording` request within the same RPC call twice. A single `Record` call will only ever correlate to a single ongoing recording. And I feel strongly that relaxing this restriction from an API design point of view would not be desirable:

To support multiple recordings within a single RPC invocation, we would have to introduce a recording identifier and then start multiplexing all messages (i.e. `Start/StopRecording` as well as the `RecordingRunningResponse` status messages) based on the identifier. We would have to introduce bookkeeping on the server-side to demultiplex all of these requests. I think this adds a lot of complexity to both the API and the implementation. However, I also believe that this complexity however is not necessary: Even with the current API, a client can simply invoke another `Record` RPC call via the _same gRPC connection_ and have gRPC/HTTP2 do the hard part of multiplexing, where it will likely do a better job than we ever could. So I really think we should stick with the "one request per `Record()`"  invocation approach we have right now.

rolinh:
> From the users perspective, if a user wants to run two concurrent recordings, the UX will be the same as it is e.g. for `tcpdump`: Start two `hubble record` commands in parallel. This has already been demonstrated to work with the current implementation.

I'm OK with this.

> From an API perspective, it is currently an error to submit a `StartRecording` request within the same RPC call twice. A single `Record` call will only ever correlate to a single recording. And I feel strongly that relaxing this restriction from an API design point of view would not be desirable:
> 
> To support multiple recordings within a single RPC invocation, (...) I really think we should stick with the "one request per `Record()`" invocation approach we have right now.

Ok, now I understand that the reason the for API to accept a stream of `RecordRequest` is only so that a stop signal can be sent to stop recording and not to support concurrent recordings. This was not clear to me initially, hence the question above.
I imagine that recording also stops if the connection is lost?

cmd/record/record.go
```diff
@@ -0,0 +1,217 @@
...
+				},
+				Include:          filters,
+				MaxCaptureLength: maxCaptureLength,
+			},
+		},
+	})
+	if err != nil {
+		return err
+	}
+	fmt.Fprintf(os.Stderr, "Started recording. Press CTRL+C to stop.\n")
```
rolinh, line 181: Could we use the command's stdout/stderr for all messages instead of `os.Stdout/os.Stderr`?

rolinh (approved):
Thanks, the PR lgtm. Not a blocker but I'd prefer that we use the command's stdout/stderr instead of `os.Stdout/os.Stderr`. This would facilitate writing tests.

gandro:
> Ok, now I understand that the reason the for API to accept a stream of `RecordRequest` is only so that a stop signal can be sent to stop recording and not to support concurrent recording requests within a single RPC call. This was not clear to me initially, hence the question above.
> I imagine that recording also stops if the connection is lost?

Correct! The reason why we have an ongoing RPC call instead of separate `Start` and `Stop` oneshot methods is exactly so we can clean up on the server-side if a client disappears without sending a stop signal.

cmd/record/record.go
gandro: So this is intentional, but I can change it: All server-side messages (Status, Result, Output) are sent to stdout, while messages from the CLI (`Press CTRL+C to stop` + `Stopping`) are sent to stderr.

This allows a user to pipe the metadata of the recording into a file or something without it being cluttered with ui prompts.

rolinh: Cobra commands have [`ErrOrStderr()`](https://pkg.go.dev/github.com/spf13/cobra#Command.ErrOrStderr) and [`OutOrStdout()`](https://pkg.go.dev/github.com/spf13/cobra#Command.OutOrStdout) methods which return an `io.Writer` suitable for err or out output. This is particularly useful for testing as one can then use `SetErr()`/`SetOut()` to set the outputs to some buffer or else that implements `io.Writer`, which allows capturing the command's output and validate it. As we don't have tests for this command yet, this is not too useful atm though, hence not a blocker to me.

gandro: Makes sense, will change it when I rebase

cmd/record/record.go
kaworu: > I'll do some digging, but I might end up only supporting Linux for now, while other OSes should just have isTTY=false, which will always work.

That is fine by me.

cmd/record/record.go
gandro: Unfortunately it turns out since I need to do a tty check, having a `io.Writer` for stdout is not sufficient. I'll leave it to `os.Stdout`/`os.Stderr` for now. If we ever need to unit test the output here, we can try to figure out what Cobra recommends here.

cmd/record/record.go
gandro: Switched to a proper TTY check on Linux now. Non-Linux platforms will currently not use the TTY logic. We can always add that later, but I don't expect this command to be used outside of the (Linux-based) Cilium pod, as it only works in Hubble local mode.

PR #532, opened 2021-04-21
Bump github.com/cilium/cilium to pull in reworked agent/debug event API

tklauser:
Update github.com/cilium/cilium to latest master to get the reworked
separate APIs for agent and debug events (see cilium/cilium#15715 and
cilium/cilium#15111).

Note that this will for now prevent the Hubble CLI from showing any
agent and debug events because the previous implementation was tightly
coupled with the GetFlows API. The actual implementation will be done as
a follow-up.


rolinh (approved):

gandro (approved):
:pray:

PR #534, opened 2021-04-23
cmd/observe: don't list agent/debug events and recorder captures in event type filter

tklauser:
As of commit 5071eca2f964 ("Bump github.com/cilium/cilium to pull in
reworked agent/debug event API"), the agent and debug event types are no
longer handled by the GetFlows API and the newly introduced recorder
capture has its own API as well, see commit df71f17a5be6 ("cmd: Add
record subcommand").

Thus, also the filtering on these in "hubble observe" which uses the
GetFlows API doesn't work. Omit these types from the --type/-t option.
Also add a test to ensure the list of supported event types are in sync
with the respective event types supported by the monitor API.

Before:

```
$ hubble observe -h
...
  -t, --type filter             Filter by event types TYPE[:SUBTYPE] ([l7 agent policy-verdict recorder drop debug capture trace])
...
```

After:

```
$ hubble observe -h
...
  -t, --type filter             Filter by event types TYPE[:SUBTYPE] ([drop capture trace l7 policy-verdict])
...
```

### DIFF
--- a/cmd/observe/observe.go
+++ b/cmd/observe/observe.go
@@ -77,11 +77,16 @@ var verdicts = []string{
 	pb.Verdict_ERROR.String(),
 }
 
-func eventTypes() (l []string) {
-	for t := range monitorAPI.MessageTypeNames {
-		l = append(l, t)
-	}
-	return
+// eventTypes are the valid event types supported by observe. This corresponds
+// to monitorAPI.MessageTypeNames, excluding MessageTypeNameAgent,
+// MessageTypeNameDebug and MessageTypeNameRecCapture. These excluded message
+// message types are not supported by observe but have separate sub-commands.
+var eventTypes = []string{
+	monitorAPI.MessageTypeNameDrop,
+	monitorAPI.MessageTypeNameCapture,
+	monitorAPI.MessageTypeNameTrace,
+	monitorAPI.MessageTypeNameL7,
+	monitorAPI.MessageTypeNamePolicyVerdict,
 }
 
 func timeFormatNameToLayout(name string) string {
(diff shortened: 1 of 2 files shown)


cmd/observe/observe.go
```diff
@@ -77,11 +77,16 @@ var verdicts = []string{
...
 
-func eventTypes() (l []string) {
-	for t := range monitorAPI.MessageTypeNames {
-		l = append(l, t)
-	}
-	return
+// eventRypes are the valid event types supported by observe. This corresponds
+// to monitorAPI.MessageTypeNames, excluding MessageTypeNameAgent,
+// MessageTypeNameDebug and MessageTypeNameRecCapture. These message types have
+// separate sub-commands.
```
kaworu, line 83: nit: typo and rephrasing:

```suggestion
// eventTypes are the valid event types supported by observe. This corresponds
// to monitorAPI.MessageTypeNames excluding MessageTypeNameAgent,
// MessageTypeNameDebug, and MessageTypeNameRecCapture. Theses excluded
// message types are not supported by observe but rather have separate sub-commands.
```

cmd/observe/observe_test.go
```diff
@@ -19,11 +19,40 @@ import (
...
+		return false
+	}
+	for k := range monitorAPI.MessageTypeNames {
+		switch k {
+		case monitorAPI.MessageTypeNameAgent,
+			monitorAPI.MessageTypeNameDebug,
+			monitorAPI.MessageTypeNameRecCapture:
+			continue
+		}
+		require.True(t, eventTypesContainsValue(k))
```
kaworu, line 52: Would [`require.Contains`](https://pkg.go.dev/github.com/stretchr/testify/require#Contains) be useful here?

kaworu (commented):
Couple of nitpicking, thanks for the cleanup PR!

cmd/observe/observe_test.go
tklauser: Thanks for the hint, changed to use `require.Contains`.

cmd/observe/observe.go
tklauser: Thanks, I've slightly adjusted the suggestion and applied it.

rolinh (approved):
Thanks!

gandro (approved):
Thanks!

kaworu (approved):

PR #540, opened 2021-04-29
Dockerfile: use alpine 3.12

aanm:
In Alpine 3.11 and 3.13, 'nslookup' exits with the error code 1 if it
can't resolve all IPs for the search list defined in /etc/resolv.conf

However, it seems that Alpine 3.10 and 3.12 are not affected by this bug
and continue return the error code 0 if at least on of the domains in
the search list is resolved into an IP address. Thus, we will use the
latest Alpine image available for the 3.12 release series.

Signed-off-by: André Martins <andre@cilium.io>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,14 @@
-FROM docker.io/library/golang:1.16.3-alpine3.13 as builder
+# Do not upgrade to alpine 3.13 as its nslookup tool returns 1, instead of 0
+# for domain name lookups.
+FROM docker.io/library/golang:1.16.3-alpine3.12@sha256:371dc6bf7e0c7ce112a29341b000c40d840aef1dbb4fdcb3ae5c0597e28f3061 as builder
 WORKDIR /go/src/github.com/cilium/hubble
 RUN apk add --no-cache git make
 COPY . .
 RUN make clean && make hubble
 
-FROM docker.io/library/alpine:3.13
+# Do not upgrade to alpine 3.13 as its nslookup tool returns 1, instead of 0
+# for domain name lookups.
+FROM docker.io/library/alpine:3.12.7@sha256:36553b10a4947067b9fbb7d532951066293a68eae893beba1d9235f7d11a20ad
 RUN apk add --no-cache bash curl jq
 COPY --from=builder /go/src/github.com/cilium/hubble/hubble /usr/bin
 CMD ["/usr/bin/hubble"]


rolinh (changes requested):
I'm not really fond of the idea of downgrading Alpine. According to the discussion on [this issue](https://github.com/gliderlabs/docker-alpine/issues/539), the problem is with the busybox implementation of `nslookup` and is not fix in either Alpine 3.11, 3.12 or 3.13, the only difference being that in Alpine 3.12, it returns error code `0` instead of `1`.
What about installing `bind-tools` to get a working version of `nslookup` instead?

```
$ docker run --rm -it alpine:3.13
Unable to find image 'alpine:3.13' locally
3.13: Pulling from library/alpine
540db60ca938: Pull complete 
Digest: sha256:69e70a79f2d41ab5d637de98c1e0b055206ba40a8145e7bddb55ccc04e13cf8f
Status: Downloaded newer image for alpine:3.13
/ # which nslookup
/usr/bin/nslookup
/ # nslookup --version
BusyBox v1.32.1 () multi-call binary.

Usage: nslookup [-type=QUERY_TYPE] [-debug] HOST [DNS_SERVER]

Query DNS about HOST

QUERY_TYPE: soa,ns,a,aaaa,cname,mx,txt,ptr,any
/ # apk add -u bind-tools
fetch https://dl-cdn.alpinelinux.org/alpine/v3.13/main/x86_64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.13/community/x86_64/APKINDEX.tar.gz
(1/17) Installing fstrm (0.6.0-r1)
(2/17) Installing krb5-conf (1.0-r2)
(3/17) Installing libcom_err (1.45.7-r0)
(4/17) Installing keyutils-libs (1.6.3-r0)
(5/17) Installing libverto (0.3.1-r1)
(6/17) Installing krb5-libs (1.18.3-r1)
(7/17) Installing json-c (0.15-r1)
(8/17) Installing libgcc (10.2.1_pre1-r3)
(9/17) Installing libstdc++ (10.2.1_pre1-r3)
(10/17) Installing libprotobuf (3.13.0-r2)
(11/17) Installing libprotoc (3.13.0-r2)
(12/17) Installing protobuf-c (1.3.3-r4)
(13/17) Installing libuv (1.40.0-r0)
(14/17) Installing xz-libs (5.2.5-r0)
(15/17) Installing libxml2 (2.9.10-r6)
(16/17) Installing bind-libs (9.16.15-r1)
(17/17) Installing bind-tools (9.16.15-r1)
Executing busybox-1.32.1-r6.trigger
OK: 19 MiB in 31 packages
/ # which nslookup
/usr/bin/nslookup
/ # nslookup -version
nslookup 9.16.15
```

rolinh (approved):
Based on an offline discussion, we decided that downgrading is the better option here.

PR #541, opened 2021-04-30
cmd: change "node list" command for "list node"

rolinh:
We recently decided to try and uniformize the Hubble CLI subcommands and
agreed on a `[verb] [noun]` approach. Given that there hasn't been any
release of Hubble CLI with the `nodes` subcommand yet, change the
`hubble nodes list` command for `hubble list nodes`.

### DIFF
--- a/cmd/list/list.go
+++ b/cmd/list/list.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package node
+package list
 
 import (
 	"github.com/cilium/hubble/cmd/common/config"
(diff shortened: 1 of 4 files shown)


michi-covalent (approved):

kaworu:
One thing to consider is that we don't mirror [`cilium node list`](https://docs.cilium.io/en/v1.9/cmdref/cilium_node_list/#cilium-node-list) after this patch.

gandro:
> One thing to consider is that we don't mirror [`cilium node list`](https://docs.cilium.io/en/v1.9/cmdref/cilium_node_list/#cilium-node-list) after this patch.

FWIW, I think that's fine. Cilium's CLI uses `noun verb` almost everywhere, whereas Hubble is doubling down on `verb noun`.

rolinh:
This raises a point about uniformity with the new cilium cli. It seems to be mixed currently (`cilium install` but `cilium hubble enable`). However, I think it's better to discuss this in our next weekly meeting or in Slack.

Issue #429, opened 2020-11-17
Add colored output to hubble observe


kaworu:
Evaluating whether adding colors to the `hubble observe` output would be a good idea
![hubble-observe-colored](https://user-images.githubusercontent.com/109415/99363120-a5740300-28b4-11eb-922a-0e035fa62ce8.png)

@joestringer noted that we used to have color support in the cilium CLI but we [removed it to drop the external dependency](https://github.com/cilium/cilium/commit/39729f31b0a06b93e22eab42e3be0840edf38180).

PR #551, opened 2021-05-11
cmd/observe: add color support

rolinh:
Add a new flag to set the color mode to the observe command. It's set to one of `auto`, `always` or `never`. Note that the color mode only works for the `dict` and `compact` mode (`table` is notably not supported).

Examples with different terminal color schemes:

image
image
image
image
image
image
image

Close: #429

### DIFF
--- a/pkg/printer/color.go
+++ b/pkg/printer/color.go
@@ -1,0 +1,109 @@
+// Copyright 2021 Authors of Hubble
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package printer
+
+import (
+	"strings"
+
+	"github.com/fatih/color"
+)
+
+type sprinter interface {
+	Sprint(a ...interface{}) string
+}
+
+type colorer struct {
+	colors  []*color.Color
+	red     sprinter
+	green   sprinter
+	blue    sprinter
+	cyan    sprinter
+	magenta sprinter
+	yellow  sprinter
+}
+
+func newColorer(when string) *colorer {
+	red := color.New(color.FgRed)
+	green := color.New(color.FgGreen)
+	blue := color.New(color.FgBlue)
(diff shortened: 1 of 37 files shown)


michi-covalent (approved):
a super basic question, but when would one use `--color=always`?

qmonnet:
> a super basic question, but when would one use `--color=always`?

When piping to something where you want the colors to be preserved, e.g. `less -R` or sometimes `grep` or you want the color markers in your output file for whatever reason.

pkg/printer/options.go
```diff
@@ -98,6 +98,20 @@ func IgnoreStderr() Option {
...
+// - "auto": color mode is enabled when the standard output is connected to a
+//   terminal.
+// - "always": color mode is enabled no matter to standard output.
+// - "never": color mode is always disabled.
+// Any other value of when means "auto", which is the default.
+// The color mode is only applied with in Dict or Compact mode. For any other
+// mode, color is always disabled.
+func WithColor(when string) Option {
+	return func(_ *Options) {
+		setColors(when)
```
glibsm, line 111: Perhaaps we should store the option here, and attach the coloring to the `Printer struct` so there is no global state?

glibsm: I wasn't specifically referring to the library usage (sometimes it forces global state like here), I was more hinting that we shouldn't add our own global `func setColors`. We shouldn't expose the innards of the library in our own code, as that gives us some more flexibility to even swap to a different library in the future without much impact.

I was thinking we can even add our own custom `struct color(er?)` and attach to options/printer to completely hide which library is even used.

---

Just out of curiosity, why is `DisableColor()` on each color not an option? All we'd really have to do is wrap all the colors into a some sort of a collection and iterate over them, no?

rolinh: Valid point. I'll rework the PR, see what I can come up with.

rolinh: I reworked the PR to avoid using globals. The new `colorer` struct could easily be adapted for a new library if need be. PTAL.

glibsm (approved):
I'm happy enough to merge as-is but left another comment for your consideration if you want to make another change.

pkg/printer/color.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func (c *colorer) disable() {
+	for _, v := range c.colors {
+		v.DisableColor()
+	}
+}
+
+func (c colorer) port(a interface{}) string {
+	return c.colors["yellow"].Sprint(a)
```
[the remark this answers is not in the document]
rolinh: > Would be great to have like colorer.green.Sprint(a), as well as them all being in a collection under the hood for other purposes

Implemented this suggestion, PTAL.

glibsm (approved):

PR #555, opened 2021-05-20
.github: Cancel outdated PR and push workflows

pchaigno:
When updating a pull request, workflows associated with the previous version continue to run. A way to prevent that and cancel outdated workflows is to use concurrency groups.

This commit handles workflows triggered by pull requests and pushes. We define each concurrency group such that they are unique to the pull request (via PR number) if triggered by event `pull_request` or to the commit (via SHA) if triggered by event `push`.


glibsm (approved):
TIL. Nice

PR #559, opened 2021-05-26
Use golangci-lint for static checks

rolinh:
See commits for details. These changes are loosely based on the changes made to cilium-cli via [#229](https://github.com/cilium/cilium-cli/pull/229).


Makefile
```diff
@@ -66,27 +69,15 @@ test:
...
-
-# Ignored staticcheck warnings:
-# - SA1019 deprecation warnings: https://staticcheck.io/docs/checks#SA1019
-# - ST1000 missing package comment: https://staticcheck.io/docs/checks#ST1000
-staticcheck:
-	./tools/staticcheck -checks="all,-SA1019,-ST1000" ./...
-
-vet:
-	go vet $$(go list ./...)
+ifneq (,$(findstring $(GOLANGCILINT_WANT_VERSION),$(GOLANGCILINT_VERSION)))
```
glibsm, line 72: should we use some sort of semver comp here, i.e. `1.40.2` should be ok too, right? Or do we specifically want the exact match

rolinh: Sure, I can drop `PATCH` from the version check.

tklauser (approved):
Nice!

PR #560, opened 2021-05-27
Makefile: Introduce GO_BUILD variable

gandro:
This allows users to overwrite the GO environment variable while still
using CGO_ENABLED=false. This for example is useful when testing
pre-releases of the Go compiler.


gandro:
> Thanks! Could we also update the `bench` and `test` targets to use `$(GO)`?

Good catch. Done and tested!

kaworu (approved):

rolinh (approved):

PR #562, opened 2021-05-31
pkg/printer: disable color output in tests

tklauser:
Depending on how `go test` is invoked, the auto-detection in the
`github.com/fatih/color` library will enable/disable color support. For
example when invoked as `go test .` in the `pkg/printer` directory,
color output will be enabled while for running `go test ./...` in the
root directory color support is disabled.

Avoid this flaky behavior by disabling color output in all printer
tests.


kaworu (approved):

rolinh (approved):
Nice catch!

PR #563, opened 2021-06-04
.github/workflows: move Go module vendoring check to build checks

tklauser:
Instead of instantiating a separate job checking out its own copy of the
code, move the module vendoring check to the build job which already has
the code checked out.


kaworu (approved):

kaworu:
@tklauser would it make sense to move theses "sanity checks" before test and build?

tklauser:
> @tklauser would it make sense to move theses "sanity checks" before test and build?

Good point, will move them.

rolinh (approved):

PR #548, opened 2021-05-10
Update Go to 1.16.4

tklauser:

PR #564, opened 2021-06-04
Update Go to 1.16.5

tklauser:


kaworu (approved):

Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Do not upgrade to alpine 3.13 as its nslookup tool returns 1, instead of 0
 # for domain name lookups.
-FROM docker.io/library/golang:1.16.4-alpine3.12@sha256:371dc6bf7e0c7ce112a29341b000c40d840aef1dbb4fdcb3ae5c0597e28f3061 as builder
+FROM docker.io/library/golang:1.16.5-alpine3.12@sha256:371dc6bf7e0c7ce112a29341b000c40d840aef1dbb4fdcb3ae5c0597e28f3061 as builder
```
rolinh, line 3: Is the checksum really identical for both versions? :wink:

rolinh: Wait, this actually builds :exploding_head:  Is the digest not validated somehow?

rolinh: OK, as it turns out, the image tag is ignored and only the digest is considered when pulling in the image. Therefore, this means that the image remains `1.16.3`.
```
$ docker run --rm  3494eb9bdfdf hubble version
hubble v0.9.0-dev@pr/update-go-1.16.5-0198d9d compiled with go1.16.3 on linux/amd64
```

tklauser: When specifying the digest, he image tag is disregarded and the build uses the image corresponding to the digest, i.e. it still used the 1.16.4 image in this case.

rolinh: Actually it still used 1.16.3 (as shown above) because the same mistake was made when upgrading to Go 1.16.4.

tklauser: Yeah, I apparently forgot to update the digest in #548.

rolinh (approved):

PR #575, opened 2021-06-22
ci: fix the go vendoring check

kaworu:
Before this patch, git diff was used to ensure that Go modules were properly vendored. While it could catch a failure to update a vendored module, it would not fail when a module was not vendored (untracked).

This patch use git status instead of git diff, effectively catching untracked vendored modules as well.

Signed-off-by: Alexandre Perrin <alex@kaworu.ch>


rolinh (approved):

tklauser (approved):
Good catch, thank you!

PR #576, opened 2021-06-22
Add CODEOWNERS

gandro:
This adds a CODEOWNERs file to the Hubble repository. It mostly just
contains a main entry matching @cilium/hubble - this is needed to set up
the GitHub code review load balancer algorithm to assign designated
reviewers. We use the same mechanism in other repositories within the
`github.com/cilium` organization.

Signed-off-by: Sebastian Wicki <sebastian@isovalent.com>

PR #577, opened 2021-06-22
CODEOWNERS: assign GH actions to github-sec team

tklauser:
Reviewed #576 a little too late: https://github.com/cilium/hubble/pull/576#pullrequestreview-689709536


gandro (approved):
Oh, good catch. I copied over the CODEOWNERs from https://github.com/cilium/image-tools/blob/master/CODEOWNERS - we probably want to use github sec there as well?

tklauser:
> Oh, good catch. I copied over the CODEOWNERs from https://github.com/cilium/image-tools/blob/master/CODEOWNERS - we probably want to use github sec there as well?

Yeah, I think we should change it there too.

rolinh (approved):

PR #581, opened 2021-06-29
ci: enable checks for missing Go documentation

rolinh:
Checks for missing Go documentation used to be covered by the now defunct project `golint`. Revive is a replacement for golint that is used by `golangci-lint`. However, `golangci-lint` [disables checks for missing Go documentation by default](https://github.com/golangci/golangci-lint/issues/456). The only way to re-enable them is apparently to suppress golangci-lint's default rules, which is what this commit does. Rules that triggered false-positive on this codebase have been manually added. The list of default exclude rules can be found by running `golangci-lint run --help`.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -34,6 +34,13 @@ linters-settings:
     go: "1.16"
 
 issues:
+  # Default rules exclude Go doc comments check, which is rather unfortunate.
+  # In order to enable Go doc checks, defaults rules have to be disabled.
+  # See https://github.com/golangci/golangci-lint/issues/456 for details.
+  exclude-use-default: false
+  exclude:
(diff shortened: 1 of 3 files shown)


.golangci.yml
```diff
@@ -34,6 +34,10 @@ linters-settings:
     go: "1.16"
 
 issues:
+  exclude-use-default: false # default rules exclude doc comments check :(
```
glibsm, line 37: i'd also include a comment pointing to https://github.com/golangci/golangci-lint/issues/456 so it's not lost in the git log if there is ever reshuffling of this file

glibsm (approved):

.golangci.yml
rolinh: Good point. Done.

nbusseneau (approved):
:shipit:

PR #582, opened 2021-07-02
Update cobra to v1.2.1 and use built-in completion command

rolinh:
Since the release of cobra v1.2.0, a completion command is automatically
created if none is provided. We therefore do not need to ship our own
anymore.

Note that this means two breaking changes:

- Calling `hubble completion` without specifying a shell name now leads
  to printing out the help message for the command. Our implementation
  used to generate completion for bash. I think that, although breaking,
  this change is for the better.
- We used to support `ps1` as an alias for `powershell` and the builtin
  subcommand does not.

What we get is more uniformity for our users. Most CLI tools using cobra
are likely to use the built-in command to generate shell completion
rather than rolling their own.


tklauser (approved):

glibsm (approved):
awesome

aanm (approved):

PR #585, opened 2021-07-13
update Go to v1.16.6, alpine to 3.14

rolinh:
Note that the Alpine had been downgraded from 3.13 to 3.12 at some point
due to a bug with busybox's nslookup implementation. This commit bumps
it to 3.14 as it seems that the bug is fixed(*) and that Alpine based
versions of Go images are only provided with Alpine 3.13 and 3.14.

(*): With Alpine 3.11 and 3.13, busybox's nslookup would return with
exit code 1 when at least one of the IP address for the search list
defined in `/etc/resolve.conf` cannot be resolved. With Alpine 3.10 and
3.12, nslookup would return exit code 0 instead. It seems that Alpine
3.14 shares the same behavior as Alpine versions 3.10 and 3.12.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,7 +24,7 @@ jobs:
         path: go/src/github.com/cilium/hubble
     - uses: actions/setup-go@v2.1.3
       with:
-        go-version: '1.16.5'
+        go-version: '1.16.6'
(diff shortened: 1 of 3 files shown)


tklauser (approved):
🚀

glibsm (approved):

PR #586, opened 2021-07-13
v0.8: bump cilium to v1.10.2, Go to v1.16.6

rolinh:
Bump Cilium to v1.10.2 and backport #585.


glibsm (approved):

michi-covalent (approved):

Dockerfile
```diff
@@ -1,14 +1,10 @@
-# Do not upgrade to alpine 3.13 as its nslookup tool returns 1, instead of 0
```
aanm, line 1: alpine 1.14 suffers from the same issue.

aanm (changes requested):

Dockerfile
rolinh: How does it affect Hubble CLI?

pchaigno: It probably doesn't, but the commit description should be updated :-)

pchaigno (approved):

Dockerfile
rolinh: As this is a backport PR, I don't want to amend the commit. I'll pull in #587 once it's merged into `master`.

pchaigno: I missed it was a backport PR because there are no references to upstream commits :scream_cat:

rolinh: Well, the process is a little different for Hubble CLI. There's v0.8 in the PR title, a `backport/0.8` label and a mention of the backported PR in the PR message. Isn't this sufficient?

pchaigno: No, no, that's fine :smile_cat: Having that information in the git history is just convenient to avoid going through GitHub all the time.

aanm (approved):

Issue #307, opened 2020-06-23
Add Hubble Binaries to Package Managers


jedsalazar:
I’m going through the kind install hubble-relay section which has users download the `darwin-amd64` image for the hubble binary on the [hubble/releases/download](https://github.com/cilium/hubble/releases) page. I’m thinking we should supply this binary in a package manager like `brew` (as well as other package managers for windows and Linux) so users automatically stay up-to-date without manually updating from the releases page. 

This will likely save us confusion in the long-run as we won't have to consistently be asking what version of the local hubble binary was installed. It's also a better operator experience as most critical binaries on an operator's machine are installed and lifecycle-managed through package managers.

zhiyanfoo:
Could you assign this issue to me? I have done some work already for homebrew. I might be available to work on other package manager but not guarantees. 

Homebrew recommends that the formula builds the target using a tarballed source. The github release of this tarballed source omit the '.git' directory, so git commands ran in the Makefile fail.

I have two PR's that solve this issue in slightly different ways and have no strong preference to which one is chosen.

https://github.com/cilium/hubble/compare/master...zhiyanfoo:makefile-command-brew-opt1
https://github.com/cilium/hubble/compare/master...zhiyanfoo:makefile-command-brew-opt2

Alternatively we could have the formula clone the repo, omitting the need to modify the makefile. Glancing through the current accepted formula's, very few repositories go through this route. 

The below formula should work nonetheless. Also available from https://github.com/zhiyanfoo/homebrew-experimental-tap.
You can try it by running 
`brew tap zhiyanfoo/experimental-tap` and `brew install hubble`.


```
require 'open3'

class Hubble < Formula
  desc "Hubble - Network, Service & Security Observability for Kubernetes using eBPF"
  homepage "https://github.com/cilium/hubble"
  url "https://github.com/cilium/hubble.git",
      tag:      "v0.8.0"
  license "Apache-2.0"
  head "https://github.com/cilium/hubble.git"

  depends_on "go" => :build

  def install
    system 'make', 'hubble'
    bin.install 'hubble'
  end

  test do
    _, _, status = Open3.capture3("#{bin}/hubble", "version")

    assert_equal 0, status
  end
end
```

So to summarize
1. Do we use the github release tarball or clone the git repo?
2. If the former which PR to modify the Makefile?
3. What other package managers would you like hubble to be available on?

rolinh:
@zhiyanfoo Awesome! I assigned the issue to you.

[...]

PR #589, opened 2021-07-15
git commands in the Makefile return the empty string if they fail.

zhiyanfoo:
In github's tarballed source code, the '.git' directory is removed, and so
git commands will fail.  This change allows compilation of hubble from
the tarballed source code.

https://github.com/cilium/hubble/issues/307

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,9 @@ INSTALL = $(QUIET)install
 BINDIR ?= /usr/local/bin
 TARGET=hubble
 VERSION=$(shell cat VERSION)
-GIT_BRANCH = $(shell which git >/dev/null 2>&1 && git rev-parse --abbrev-ref HEAD)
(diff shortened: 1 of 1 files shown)


zhiyanfoo:
Tested by moving the .git directory into a random subdirectory and seeing `make hubble` still worked.

rolinh (approved):
> Tested by moving the .git directory into a random subdirectory and seeing `make hubble` still worked.

I can confirm that the version is also correctly reported:
```
$ rm -rf .git
$ make
CGO_ENABLED=0 go build  -ldflags "-w -s -X 'github.com/cilium/hubble/pkg.GitBranch=' -X 'github.com/cilium/hubble/pkg.GitHash=' -X 'github.com/cilium/hubble/pkg.Version=0.9.0-dev'" -o hubble
$ ./hubble version
hubble v0.9.0-dev compiled with go1.16.5 on linux/amd64
```

Note that the change is mostly cosmetic.

PR #596, opened 2021-07-22
flow: added policy name and audit_type to flow json format

MahaPanch:
update stable.txt to v0.8.1

Signed-off-by: Mahadev Panchal <mahavishwa.07@gmail.com>


MahaPanch:
Need to take this change in feature

PR #607, opened 2021-08-11
cmd/record: Add stop condition flags 

gandro:
This adds three new flags to the `hubble record` subcommand:

 - `packet-limit`: Stops the recording after a specified number of
   packets have been recorded.
 - `bytes-limit`: Stops the recording after a specified number of
   bytes have been recorded.
 - `time-limit`: Stops the recording after a specified duration
   has elapsed.

The first limit to hit will stop the recording. If none of the three
flags are specified, then the recording needs to be stopped manually.

These flags are only supported by Cilium master at the moment, see
cilium/cilium#16473 for the API implementation.


michi-covalent (approved):
🚢

cmd/record/record.go
```diff
@@ -80,6 +86,10 @@ protocols are TCP, UDP and ANY.`,
 	recorderFlags.StringVar(&fileSinkPrefix, "file-prefix", "hubble", "File prefix of the resulting pcap file")
 	recorderFlags.Uint32Var(&maxCaptureLength, "max-capture-len", 0, "Sets the maximum capture length (zero for full capture)")
 
+	recorderFlags.Uint64Var(&packetLimit, "packet-limit", 0, "Sets a limit on how many packets to capture on each node")
+	recorderFlags.Uint64Var(&bytesLimit, "bytes-limit", 0, "Sets a limit on how many bytes to capture on each node")
+	recorderFlags.DurationVar(&timeLimit, "time-limit", 0, "Sets a limit on how long to capture on each node")
```
glibsm, line 91: does this do the relative nifty things like `10s`, `5m` etc?

rolinh (approved):

PR #614, opened 2021-08-27
cmd/status: add support for multiple output formats

rolinh:
The existing behavior is maintained and translated to the "compact" format, which is the default. In addition, new output format are supported: dict, json and table. These formats are consistent with the formatting options of the `observe` command.

### DIFF
--- a/pkg/printer/printer_test.go
+++ b/pkg/printer/printer_test.go
@@ -1090,3 +1090,157 @@ Jan  1 00:20:34.567   k8s1   cilium-test/pod-to-a-denied-cnp-75cb89dfd-vqhd9 (ID
 		})
 	}
 }
+
+func TestPrinter_WriteServerStatusResponse(t *testing.T) {
+	buf := bytes.Buffer{}
+	ss := &observerpb.ServerStatusResponse{
+		NumFlows:  2031,
+		MaxFlows:  4095,
+		SeenFlows: 2348885,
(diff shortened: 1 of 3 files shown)


kaworu (approved):

pkg/printer/printer_test.go
```diff
@@ -1090,3 +1090,157 @@ Jan  1 00:20:34.567   k8s1   cilium-test/pod-to-a-denied-cnp-75cb89dfd-vqhd9 (ID
...
+		MaxFlows:            8190,
+		SeenFlows:           2771,
+		UptimeNs:            301515181665,
+		Version:             "hubble-relay v1.10.3+g4145278",
+		NumConnectedNodes:   &wrapperspb.UInt32Value{Value: 2},
+		NumUnavailableNodes: &wrapperspb.UInt32Value{Value: 0},
+	}
+	type args struct {
+		ss *observerpb.ServerStatusResponse
+	}
```
glibsm, line 1114: kind of an odd contraption. looks like there may have been more args in there at one point

glibsm (approved):

pkg/printer/printer_test.go
rolinh: Yeah, I just re-used the style of the existing tests. I wouldn't have used an `args` struct as there's only one argument anyways.

PR #619, opened 2021-09-07
[v0.8] go.mod, vendor: bump cilium to v1.10.4

tklauser:


kaworu (approved):

glibsm:
@tklauser Is this something that you have automated? Judging by the vendor diff, this is a no-op from the hubble perspective so just curious as to the motivation.

tklauser:
@glibsm this isn't automated (yet). Not sure whether we could/want to enable something like dependabot for release branches.

The motivation is to avoid transitive dependencies on older Cilium releases, e.g. in cilium-cli where we always bump to the latest release. There has been an issue with a retracted version of a package that cilium depended on (though in a non-release version) which was preventing module updates in cilium-cli.

PR #625, opened 2021-09-13
ci: bump Go to 1.17 for golangci-lint

kaworu:
Specify go version to be used by `stylecheck` along with some cosmetic changes on the way.

See also https://github.com/cilium/hubble/pull/620


rolinh (approved):
Nice catch!

tklauser (approved):
Good catch :+1:

PR #626, opened 2021-09-16
printer: Add missing verdicts

pchaigno:
This pull request adds some missing verdicts to the CLI. In summary, commits:
1. updates the cilium/cilium dependency to pull in https://github.com/cilium/cilium/pull/17411.
2. adds the missing `Verdict_AUDIT` and `Verdict_ERROR` cases.
3. adds the new `Verdict_REDIRECTED` case.

See commits for details.

Marked for v0.8 backports, but not sure about the backport policy.


rolinh:
> Marked for v0.8 backports, but not sure about the backport policy.

I see that the related Cilium PR is marked for backport to v1.10. Given that Hubble v0.8 supports Cilium v1.10, it's appropriate to also label this PR for backporting.

cmd/observe/flows.go
```diff
@@ -46,6 +46,8 @@ import (
 var verdicts = []string{
 	flowpb.Verdict_FORWARDED.String(),
 	flowpb.Verdict_DROPPED.String(),
+	flowpb.Verdict_AUDIT.String(),
```
glibsm, line 49: This is not something that was added in the PR. I wonder why it was never added

pchaigno: > This is not something that was added in the PR.

I'm not sure what you mean. This was added on purpose by the second commit in this pull request.

> I wonder why it was never added

I assumed it was an oversight. I don't see a reason we would want to keep this hidden.

tklauser (approved):

cmd/observe/flows.go
glibsm: I didn't look at the commit basis at first. Yes, this is fine to commit as is, I just thought it was squashed all in one.

glibsm (approved):

pchaigno:
Rebased. All review requests are covered and tests are passing. Marking ready to merge.

rolinh (approved):

PR #627, opened 2021-09-16
github: Add "Image Release Build" workflow 

gandro:
This adds a workflow to build the release images via GitHub actions,
instead of relying on the quay.io autobuilder.

The quay.io autobuilder often fails due to DockerHub's rate
limiting on the golang builder base image. By using our own
Docker credentials, we can avoid hitting the rate limit.

The workflow heavily based on the one found in `cilium/cilium`, with
only minor modifications (i.e. only building a single image, only
pushing to quay.io) and is similar to the one used by Hubble UI.


gandro:
Note: I tried to add support for arm64, but it seems like the arm64 alpine base images have some issues when running on x86-bsaed GitHub Actions: https://gitlab.alpinelinux.org/alpine/aports/-/issues/12406

gandro:
Successful run demonstrating credentials and build works (with `push: false`) https://github.com/cilium/hubble/actions/runs/1253368334

nbusseneau (approved):
LGTM overall, did you try it out manually via triggers other than `push`?

gandro:
> LGTM overall, did you try it out manually via triggers other than `push`?

Yes. This is the commit which led to the successful run that I linked above:
https://github.com/cilium/hubble/commit/52c8c7bc045b837246ee7b1c7492f6153c86cac1

nbusseneau:
Ah sorry I missed the comment linking to a successful run 🙈

nebril (approved):

PR #629, opened 2021-09-22
printer: group digits by 3 for flow counters and make uptime human-readable 

rolinh:
Large numbers are hard to read by a human due to the number of consecutive digits. Grouping digits by 3 improves readability of large numbers such as the number of flows when using a large Hubble ring buffer and/or going through Hubble Relay. This patch changes how flow counters are displayed on `hubble status` command results such that digits are grouped by 3, separated by a comma.

An example is probably worth a thousand words. The following `hubble status` result:
```
$ hubble status
Current/Max Flows: 1966080/1966080 (100.00%)
Flows/s: 845.46
```

Is turned to the following with this patch:
```
$ hubble status
Current/Max Flows: 1,966,080/1,966,080 (100.00%)
Flows/s: 845.46
```

I've been using `hubble` with this patch for the past few days and found it easily more readable for large numbers. Feedback welcome though!

EDIT: pushed another commit to also format uptime as a human readable string rather than the raw number of nanoseconds elapsed:
```
$ hubble status -o dict
          NUM FLOWS: 8,190
          MAX FLOWS: 8,190
         SEEN FLOWS: 618,983
             UPTIME: 13h33m2.571715601s
NUM CONNECTED NODES: 2
 NUM UNAVAIL. NODES: 0
            VERSION: hubble-relay v1.10.4
```


pkg/printer/formatter.go
```diff
@@ -0,0 +1,40 @@
...
+
+package printer
+
+import (
+	"strconv"
+)
+
+// uint64Grouping formats n by grouping digits by 3, separated by a comma
+// (e.g. 1000000 is formatted as 1,000,000).
+func uint64Grouping(n uint64) string {
```
glibsm, line 23: dibs on never having to maintain this function

pkg/printer/printer_test.go
```diff
@@ -1206,9 +1206,9 @@ Connected Nodes: 2/2`,
 			args:    args{ss},
 			wantErr: false,
 			expected: `
-          NUM FLOWS: 2031
-          MAX FLOWS: 4095
-         SEEN FLOWS: 2348885
+          NUM FLOWS: 2,031
+          MAX FLOWS: 4,095
+         SEEN FLOWS: 2,348,885
           UPTIME NS: 301515181665
```
glibsm, line 1212: do we not want it here either?

glibsm (commented):
I think we should consider either a flag to turn this off, or a flag to turn this on in the first place and default to as-is. What do you think?

rolinh:
> I think we should consider either a flag to turn this off, or a flag to turn this on in the first place and default to as-is. What do you think?

That's a good point. A flag is probably a good idea indeed. In an offline discussion, @aanm suggested formatting large numbers like this `2M (1966080)` which I think is a good idea as it would avoid having any discussion about locale support (some countries use some character other than `,` for  thousands separators such as `.` or `'`) and we'd get an approximate number that's easy to read by a human as well as the exact value.

pkg/printer/printer_test.go
rolinh: Yeah, I was wondering about this one. We could even format this as a readable duration no? Something like `1h20` etc.

glibsm:
> > I think we should consider either a flag to turn this off, or a flag to turn this on in the first place and default to as-is. What do you think?
> 
> That's a good point. A flag is probably a good idea indeed. In an offline discussion, @aanm suggested formatting large numbers like this `2M (1966080)` which I think is a good idea as it would avoid having any discussion about locale support (some countries use some character other than `,` for thousands separators such as `.` or `'`) and we'd get an approximate number that's easy to read by a human as well as the exact value.

Seeing 2M/2M seems a little too vague to me, I think

If we want locale there is `golang.org/x/text/message` we can pull in, although I've never tried to determine the system language. Surely there is some way

```
package main

import (
    "golang.org/x/text/language"
    "golang.org/x/text/message"
)

func main() {
    p := message.NewPrinter(language.English)
    p.Printf("%d\n", 1000)

    // Output:
    // 1,000
}
```

kaworu (approved):
Comma is fine by me. I see little value of ocalization without translation, just my 2 cents.

pkg/printer/printer_test.go
rolinh: Pushed a commit to implement this, PTAL.

pkg/printer/formatter.go
```diff
@@ -38,3 +39,13 @@ func uint64Grouping(n uint64) string {
 		}
 	}
 }
+
+// formatDurationNS formats a duration expressed in nanosecond to a human
+// readable duration. For example, 100_000_000_000 is formatted as 1m40s.
+func formatDurationNS(ns uint64) string {
+	if ns > (1<<63 - 1) {
```
kaworu, line 46: nit

```suggestion
	if ns > math.MaxInt64 {
```

kaworu (approved):

rolinh:
> I think we should consider either a flag to turn this off, or a flag to turn this on in the first place and default to as-is. What do you think?

I've given this more thoughts and think that a flag is actually not necessary. If the output is to be processed by a human, then I think the more readable, the better. If the output is not meant to be processed by a human, then I think that `--output json` should be used instead.

I've also pushed another commit to make the uptime duration human-readable. PTAL.

pkg/printer/formatter.go
jrfastab: Could this use the text/messages lib?

https://play.golang.org/p/BPPSSpNqgH2

pkg/printer/formatter.go
```diff
@@ -38,3 +40,13 @@ func uint64Grouping(n uint64) string {
 		}
 	}
 }
+
+// formatDurationNS formats a duration expressed in nanosecond to a human
+// readable duration. For example, 100_000_000_000 is formatted as 1m40s.
+func formatDurationNS(ns uint64) string {
```
jrfastab, line 46: Could we turn ns into a Duration and use normal %d printer?

rolinh: I'm not sure I understand your comment as this is basically what this function does (see the return statement). The only different is that it accounts for cases where we have a uint64 that would overflow an int64 (unlikely to happen in practice but probably best to handle it) which is the base time of a time.Duration.

pkg/printer/formatter.go
rolinh: I'm not too worried about this. It's unlikely that we'll touch this function and it's covered by tests. When the need arises to handle floats, ints etc I think it'll be time to pull in `golang.org/x/text/message`. until then I think this function is good enough.

glibsm (approved):

PR #653, opened 2021-12-06
Update Go to 1.17.4 and alpine to 3.15

tklauser:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM docker.io/library/golang:1.17.3-alpine3.14@sha256:ea5d6a7cf667df5041c09dc3741fc091fc07e6c8f996b580c1161d44313358b4 as builder
+FROM docker.io/library/golang:1.17.4-alpine3.14@sha256:c1261b2f0b62eb03d4d17b9b491d7029ec85e3e384d03ab9e0e24add03388f8f as builder
```
rolinh, line 1: Should we also bump Alpine to 3.15 while we're at it?

tklauser:
Not sure about the unit test failures: https://github.com/cilium/hubble/runs/4428100765?check_suite_focus=true

The tests passed locally, but in CI the non-breaking space (`\u00a0`) introduced in 771bb4ad3441f211f72d7aa9d0f118b0831ad72b seems to break the test. /cc @gandro

Dockerfile
tklauser: Bumped all alpine-based images to 3.15 in the latest revision.

rolinh (approved):

gandro:
> The tests passed locally, but in CI the non-breaking space (\u00a0) introduced in 771bb4a seems to break the test. /cc @gandro

Odd. I had to introduce the nbsp due to the newly generated protobuf go bindings in the last Cilium update. Seems like updating Go makes this go away again? What version of Go are you using locally?

tklauser:
> > The tests passed locally, but in CI the non-breaking space (\u00a0) introduced in [771bb4a](https://github.com/cilium/hubble/commit/771bb4ad3441f211f72d7aa9d0f118b0831ad72b) seems to break the test. /cc @gandro
> 
> Odd. I had to introduce the nbsp due to the newly generated protobuf go bindings in the last Cilium update to actually fix the test. Seems like updating Go makes this go away again? What version of Go are you using locally?

I'm using Go 1.17.4 locally, as in this PR.

Digging a bit deeper, it looks like the use of `U+00a0` is intentionally non-deterministic to avoid users from matching on the error string :grin:

https://github.com/protocolbuffers/protobuf-go/blob/26e8bcb3c743193558d1a0ff540c9e05f999267d/internal/errors/errors.go#L26-L34

I guess we should probably just match the parts we are interested int.

tklauser:
> I guess we should probably just match the parts we are interested in.

Sent https://github.com/cilium/hubble/pull/655 with a suggested fix.

gandro:
> Digging a bit deeper, it looks like the use of `U+00a0` is intentionally non-deterministic to avoid users from matching on the error string grin
> 
> https://github.com/protocolbuffers/protobuf-go/blob/26e8bcb3c743193558d1a0ff540c9e05f999267d/internal/errors/errors.go#L26-L34
> 
> I guess we should probably just match the parts we are interested in.

:see_no_evil: Really nice find! Thanks for digging in!

nbusseneau (approved):

PR #675, opened 2022-01-24
.github: let dependabot ignore Cilium dependency

tklauser:
To properly update Cilium, the replace directive in go.mod needs to be
updated as well, something dependabot can't do.


rolinh (approved):
Thanks! I somehow thought we already had dependabot to ignore the cilium/cilium dep but I'm probably confusing with cilium-cli.

Issue #677, opened 2022-01-28
Trying to set a hubble parameter results in ENOENT.


sdake:
sdake@devenv ~ [127]> ./hubble config set tls true
open /home/sdake/.config/hubble/config.yaml: no such file or directory
sdake@devenv ~ [1]> ./hubble config get tls
false

I could probably create this file, which is what I'm going to do next. The help included within the binary indicates that config.yaml is optional. The loading order also specifies this file as optional.

Cheers,
-steve

rolinh:
Hi Steven, thanks for reporting the issue. I can reproduce and my understanding is that if the file is missing, it's not a problem but if the _directory_ is missing, then one gets the error that you report.

```
$ hubble config set tls true
open /home/robin/.config/hubble/config.yaml: no such file or directory
$ file /home/robin/.config/hubble/config.yaml 
/home/robin/.config/hubble/config.yaml  cannot open `/home/robin/.config/hubble/config.yaml ' (No such file or directory)
$ mkdir /home/robin/.config/hubble
$ hubble config set tls true
$ file /home/robin/.config/hubble/config.yaml
/home/robin/.config/hubble/config.yaml: ASCII text
```

PR #684, opened 2022-02-24
cmd/config: ensure that the configuration directory exist

rolinh:
`hubble config set ...` fails to save the configuration file if the
configuration directory doesn't already exist thus requiring the user to
create it beforehand. To address this issue, this commit ensures that
the configuration directory exists when `config set` is invoked by
creating it when necessary.

Closes: #677

### DIFF
--- a/cmd/config/set.go
+++ b/cmd/config/set.go
@@ -18,6 +18,7 @@ import (
 	"encoding/csv"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


kaworu (approved):
Not sure wether `newFileOnlyViper` is the right place for this logic (vs for example set's command `PreRunE`), but LGTM.

PR #692, opened 2022-03-11
ipcache: Accept metadata from any source

michi-covalent:
The metadata source field in ipcache is not necessarily set to "k8s" in
newer versions of Cilium. This patch updates the ipcache initialization
logic to accept pod information from any metadata source.

Ref: https://github.com/cilium/cilium/issues/15698
Ref: https://github.com/cilium/cilium/pull/15706

### DIFF
--- a/pkg/ipcache/ipcache.go
+++ b/pkg/ipcache/ipcache.go
@@ -21,7 +21,6 @@ import (
 
 	"github.com/cilium/cilium/api/v1/models"
 	"github.com/cilium/cilium/pkg/identity"
-	"github.com/cilium/cilium/pkg/source"
 )
 
 // IPIdentity contains the data associated with an IP address
@@ -160,7 +159,7 @@ func (ipc *IPCache) InitializeFrom(entries []*models.IPListEntry) error {
 		}
 
 		var ns, pod string
-		if e.Metadata != nil && e.Metadata.Source == string(source.Kubernetes) {
+		if e.Metadata != nil {
 			ns = e.Metadata.Namespace
 			pod = e.Metadata.Name
 		}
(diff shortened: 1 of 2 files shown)


jrfastab:
ACK and Tested in a cluster here.

joestringer (approved):

pkg/ipcache/ipcache.go
```diff
@@ -160,7 +160,8 @@ func (ipc *IPCache) InitializeFrom(entries []*models.IPListEntry) error {
 		}
 
 		var ns, pod string
-		if e.Metadata != nil && e.Metadata.Source == string(source.Kubernetes) {
+		if e.Metadata != nil &&
+			(e.Metadata.Source == string(source.Kubernetes) || e.Metadata.Source == string(source.CustomResource)) {
```
joestringer, line 164: I'm not sure about the original rationale for filtering the source here, we could probably just check `e.Metadata != nil`. But maybe @gandro or others have some idea. In particular, this version as-is would probably break in an environment where Cilium is deployed with etcd as the backend for sharing ipcache between nodes.

gandro: The original rationale was that the source would tell you what kind of `Name` the name refers to. `Source: kubernetes`would mean that `Name` refers to a pod name (and not e.g. a docker container name).

But we never added other kinds of sources. Instead, we repurposed the `Source` field to reflect the IPCache entry source to aid with debug-ability: https://github.com/cilium/cilium/pull/15706

Thus I agree, let's remove the source check completely.

michi-covalent: fixed!

PR #707, opened 2022-04-21
Add support for less granular time formats

rolinh:


pkg/time/time.go
```diff
@@ -21,6 +21,15 @@ import (
 )
 
 const (
+	// YearMonthDay is a time format similar to RFC3339 with day granularity
+	// instead of second.
+	YearMonthDay = "2006-01-02"
+	// YearMonthDayHour is a time format similar to RFC3339 with hour
+	// granularity instead of second.
+	YearMonthDayHour = "2006-01-02T15Z"
```
kaworu, line 29: If the goal of the trailing Z is to allow specifiying the tz, it should be:
```suggestion
	YearMonthDayHour = "2006-01-02T15Z07:00"
```

pkg/time/time.go
```diff
@@ -21,6 +21,15 @@ import (
...
+	// YearMonthDayHourMinute is a time format similar to RFC3339 with minute
+	// granularity instead of second.
+	YearMonthDayHourMinute = "2006-01-02T15:04Z"
```
kaworu, line 32: ditto
```suggestion
	YearMonthDayHourMinute = "2006-01-02T15:04Z07:00"
```

pkg/time/time_test.go
```diff
@@ -43,22 +43,36 @@ func TestFromString(t *testing.T) {
...
+			expected: "2019-06-30T00:00:00Z",
+		},
+		{
+			input:    "2019-06-30T18Z",
+			expected: "2019-06-30T18:00:00Z",
+		},
+		{
+			input:    "2019-06-30T18:45Z",
+			expected: "2019-06-30T18:45:00Z",
+		},
```
kaworu, line 57: ditto
```suggestion
		},
		{
			input:    "2019-06-30T18:45+02:00",
			expected: "2019-06-30T16:45:00Z",
		},
```

cmd/observe/flows.go
```diff
@@ -349,23 +395,36 @@ more.`,
...
+  compact:  Compact output
+  dict:     Each flow is shown as KEY:VALUE pair
+  json:     JSON encoding
+  jsonpb:   Output each GetFlowResponse according to proto3's JSON mapping
+  table:    Tab-aligned columns
 `)
 	formattingFlags.BoolVarP(&formattingOpts.nodeName, "print-node-name", "", false, "Print node name in output")
 	formattingFlags.StringVar(
 		&formattingOpts.timeFormat, "time-format", "StampMilli",
 		fmt.Sprintf(`Specify the time format for printing. This option does not apply to the json and jsonpb output type. One of:
```
kaworu, line 407: I don't think the formats introduced by this PR make a lot of sense for output. On the other hand, it's probably less confusing that way because of the symmetry with `--since` and `--until`, so no objection here.

kaworu (changes requested):

cmd/observe/flows.go
rolinh: I agree. My take is that it would be more code to disallow this format for output and as it also doesn't hurt to have it, I thought it's fine.

michi-covalent (approved):
🚢

kaworu (approved):
Thanks @rolinh, patch LGTM now.

PR #717, opened 2022-05-12
printer: Display security identity in compact output

gandro:
This commit changes the default compact mode output to always display
the Cilium security identity. This is very useful when debugging policy
related issues, as those often occur when an endpoint does not have the
expected security identity assigned to it.

Screenshot below. Note the magenta `(identity)` fields which are added by this PR.

image

There is no flag to disable those identities in compact mode. Users who rely on stable output are asked to use the `json` or `jsonpb` output.


michi-covalent (approved):
❤️ this will make troubleshooting much easier.

PR #715, opened 2022-05-11
Update Go to 1.18.2

tklauser:

PR #723, opened 2022-05-16
Dockerfile: fix golang image name to v1.18.2

kaworu:
Missed by https://github.com/cilium/hubble/pull/715, I don't understand how it can work with the mismatching version/sha :shrug:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM docker.io/library/golang:1.18.1-alpine3.15@sha256:e6b729ae22a2f7b6afcc237f7b9da3a27151ecbdcd109f7ab63a42e52e750262 as builder
+FROM docker.io/library/golang:1.18.2-alpine3.15@sha256:e6b729ae22a2f7b6afcc237f7b9da3a27151ecbdcd109f7ab63a42e52e750262 as builder
 WORKDIR /go/src/github.com/cilium/hubble
 RUN apk add --no-cache git make
 COPY . .


joamaki:
Looks like if you specify the sha256 it doesn't care about the label anymore:
```
cilium % docker run -it --rm docker.io/library /golang:whatever@sha256:e6b729ae22a2f7b6afcc237f7b9da3a27151ecbdcd109f7ab63a42e52e750262 
/go #
```

tklauser:
> Missed by #715, I don't understand how it can work with the mismatching version/sha shrug

If a sha is given, Docker only considers that one and ignores the tag. The tag can have any arbitrary value.

joamaki (approved):
Looks good, the tag now matches with sha256:
```
cilium % docker run -it --rm docker.io/library/golang:1.18.2-alpine3.15                                                                                   
Unable to find image 'golang:1.18.2-alpine3.15' locally
1.18.2-alpine3.15: Pulling from library/golang
Digest: sha256:e6b729ae22a2f7b6afcc237f7b9da3a27151ecbdcd109f7ab63a42e52e750262
Status: Downloaded newer image for golang:1.18.2-alpine3.15
```

kaworu:
Oh I see I guess it make sense since tags can be updated to point to another sha, which is the reason why we're pinning images in the first place.

But then if there is a "vulnerable" version `golang:X@sha256:FOO` and patched version `golang:Y@sha256:BAR`, it's possible to make someone think `golang:Y@sha256:FOO` is not vulnerable but it is correct (because I guess humans mostly look at the `Y` version rather than the sha)? Not ideal.

Issue #501, opened 2021-03-03
Document pod/service filter prefix behavior


rolinh:
This is the current help message for the pod and service filters:
```
      --from-pod filter         Show all flows originating in the given pod name ([namespace/]<pod-name>). If namespace is not provided, 'default' is used
      --from-service filter     Show all flows originating in the given service ([namespace/]<svc-name>). If namespace is not provided, 'default' is used
      --pod filter              Show all flows related to the given pod name ([namespace/]<pod-name>). If namespace is not provided, 'default' is used
      --service filter          Show all flows related to the given service ([namespace/]<svc-name>). If namespace is not provided, 'default' is used
      --to-pod filter           Show all flows terminating in the given pod name ([namespace/]<pod-name>). If namespace is not provided, 'default' is used
      --to-service filter       Show all flows terminating in the given service ([namespace/]<svc-name>). If namespace is not provided, 'default' is used
```

It appears that `<pod-name>` and `<service-name>` are treated as prefixes for the actual pod or service name. This allows performing queries like `hubble observe --pod cilium` to get all flows related to pods which name starts with `cilium`. This behavior is currently undocumented.

rolinh:
Hey @slayer321,

What is undocumented is that the these filters are not _exact_ match, they are _prefix_ match. In other words, `hubble observe --pod kube-system/cilium` will match _every_ cilium pod as their names start with `cilium` although the full name is something like `cilium-phzv7`. Is this explanation more clear?

slayer321:
Yeah, Now I'm able to understand it. Will raise a PR now :)

PR #733, opened 2022-06-03
docs: Document pod/service filter prefix behavior

slayer321:
Document pod/service filter prefix behavior

Fixes: #501 
Signed-off-by: slayer321 <sachin.maurya7666@gmail.com>

### DIFF
--- a/cmd/observe/flows.go
+++ b/cmd/observe/flows.go
@@ -316,13 +316,13 @@ more.`,
 
 	filterFlags.Var(filterVar(
 		"from-pod", ofilter,
-		"Show all flows originating in the given pod name ([namespace/]<pod-name>). If namespace is not provided, 'default' is used"))
+		"Show all flows originating in the given pod name prefix([namespace/]<pod-name>). If namespace is not provided, 'default' is used"))
 	filterFlags.Var(filterVar(
 		"pod", ofilter,
-		"Show all flows related to the given pod name ([namespace/]<pod-name>). If namespace is not provided, 'default' is used"))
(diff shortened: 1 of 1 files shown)


rolinh (changes requested):
According to #501, there are 6 docstrings to fix (`--from-pod`, `--pod`, `--to-pod` and `--from-service`, `--service`, `--to-service`), not just `--pod` and `--service`. I also find the description too verbose. I think that something along the line of the following would be enough:
```
      --from-pod filter         Show all flows originating in the given pod name prefix ([namespace/]<pod-name>). If namespace is not provided, 'default' is used.

```

cmd/observe/flows.go
```diff
@@ -316,13 +316,13 @@ more.`,
 
 	filterFlags.Var(filterVar(
 		"from-pod", ofilter,
-		"Show all flows originating in the given pod name ([namespace/]<pod-name>). If namespace is not provided, 'default' is used"))
+		"Show all flows originating in the given pod name prefix([namespace/]<pod-name>). If namespace is not provided, 'default' is used"))
 	filterFlags.Var(filterVar(
 		"pod", ofilter,
-		"Show all flows related to the given pod name ([namespace/]<pod-name>). If namespace is not provided, 'default' is used"))
+		"Show all flows originating in the given pod name prefix ([namespace/]<pod-name>). If namespace is not provided, 'default' is used."))
```
rolinh, line 322: I think `related to` was correct here as this flag is about flows that are either from or to the given pod.
```suggestion
		"Show all flows related to the given pod name prefix ([namespace/]<pod-name>). If namespace is not provided, 'default' is used."))
```

rolinh:
Thanks @slayer321. Could you please squash your commits before we merge the PR?

PR #734, opened 2022-06-03
compact: Use "ID" for security identity prefix

michi-covalent:
It's shorter than "identity", and still conveys enough information
about what these integers mean.


gandro (approved):

gandro:
Changing the release notes to misc, since the output it changes is not released yet.

PR #738, opened 2022-06-07
cli: Deprecate `-o json`, recommend `-o jsonpb` instead

gandro:
The `-o json` output had been added to Hubble very early, when we
assumed that all relevant information would be contained in the `Flow`
protobuf message. Hubble has however since gained the `-o jsonpb`
output, which contains additional information such as the node name
which produced the network flow.

More importantly, `-o json` cannot be piped into `hubble observe` for
additional filtering, therefore further limiting its usability. To avoid
users accidentally using `-o json` instead of the likely more correct
`-o jsonpb`, this commit deprecates `-o json` and prints a warning to
stderr whenever it is used.

This allows us to fully remove `-o json` in a future release.


rolinh (approved):

pkg/printer/options.go
```diff
@@ -52,6 +54,7 @@ type Option func(*Options)
 
 // JSON encoded output from the printer.
 func JSON() Option {
+	fmt.Fprintf(os.Stderr, "WARNING: The --output=json option has been deprecated. Use --output=jsonpb instead\n")
```
kaworu, line 57: nit
```suggestion
	fmt.Fprintln(os.Stderr, "WARNING: The --output=json option has been deprecated. Use --output=jsonpb instead")
```

kaworu (approved):

michi-covalent (approved):

pchaigno (approved):

Issue #462, opened 2021-01-11
Document subtypes for the --type flag


kaworu:
`hubble --help` shows only the types for the `--type`flag (i.e. `l7 agent policy-verdict drop debug capture trace`) while it also accept a subtype part.

_Originally posted by @kaworu in https://github.com/cilium/cilium/pull/14477#discussion_r551214269_

PR #744, opened 2022-06-13
cmd/observe: document subtypes and add completion for subtypes

rolinh:
The `--type` flag optionally accepts subtypes to be specified (currently
for the type `trace` only). This patch documents which types accept
subtypes and list all available subtypes. In addition, it improves flag
completion such that subtypes are also offered.

```
$ hubble help observe
(...)
  -t, --type filter             Filter by event types TYPE[:SUBTYPE], available types and subtypes:
                                TYPE             SUBTYPE
                                capture          n/a
                                drop             n/a
                                l7               n/a
                                policy-verdict   n/a
                                trace            from-endpoint
                                                 to-endpoint
                                                 from-host
                                                 to-host
                                                 from-network
                                                 to-network
                                                 from-overlay
                                                 to-overlay
                                                 from-proxy
                                                 to-proxy
                                                 from-stack
                                                 to-stack
(...)
```
image

Note: I think `trace` is the only type that currently accepts subtypes. Please, correct me if I'm wrong :slightly_frowning_face: 

Closes: #462


gandro (approved):
Awesome!

PR #747, opened 2022-06-21
Update Cobra to v1.5.0 and use new group flags feature

rolinh:
Cobra v1.5.0 is now [released](https://github.com/spf13/cobra/releases/tag/v1.5.0) and with it comes a new API called [flag groups](https://github.com/spf13/cobra/blob/master/user_guide.md#flag-groups) that allows to mark flags as mutually exclusive or required together. 

~This PR gives it a go~ but other than saving a couple of lines of code, I'm not convince by the change from a user perspective.

Current output:
```
$ hubble observe --first 42 --all
cannot set both --first and --all
```

With this patch:
```
$ hubble observe --first 42 --all
if any flags in the group [first all] are set none of the others can be; [all first] were all set
```

Note that this new Cobra release also brings another new feature called [active help](https://github.com/spf13/cobra/blob/master/active_help.md). I'm not yet sure how it'd be best used in the context of the Hubble CLI.

EDIT: I ended up not opting for the new flag groups feature and removed the commit.


kaworu:
> ```
> $ hubble observe --first 42 --all
> if any flags in the group [first all] are set none of the others can be; [all first] were all set
> ```

I find this error message more confusing than before the patch, unfortunately it cannot be customized: https://github.com/spf13/cobra/blob/5f2ec3c897155c3346e77430932e3966e5cfa648/flag_groups.go#L160. Let's see what the other people think.

Also side question: aren't `--all` and `--last` also mutually exclusive?

tklauser:
> > ```
> > $ hubble observe --first 42 --all
> > if any flags in the group [first all] are set none of the others can be; [all first] were all set
> > ```
> 
> I find this error message more confusing than before the patch, unfortunately it cannot be customized: https://github.com/spf13/cobra/blob/5f2ec3c897155c3346e77430932e3966e5cfa648/flag_groups.go#L160. Let's see what the other people think.

I agree with Alex that the previous error message is more distinct and easier to read. As long as the message cannot be customized, I think we should stick to the old way of checking mutually exclusive flags.

rolinh:
> Also side question: aren't `--all` and `--last` also mutually exclusive?

Indeed. I added a commit to handle this case.

As hinted in the PR description, I also agree that the new error message is not very clear nor succinct. I removed the commit that switched to using new flag groups.

tklauser (approved):
Thanks

PR #751, opened 2022-06-22
Makefile: Run release build as regular user

gandro:
This fixes an issue where `make release` would fail to build the release
binaries, because `go build` would fail with `error obtaining VCS
status: exit status 128`.

This happens because `go build` in Go v1.18 and newer is invoking `git`
as part of the build process. However, due to
[CVE-2022-24765](https://github.blog/2022-04-12-git-security-vulnerability-announced/),
git v2.35.2 now requires that the current git directory for most
commands is owned by the user with which the `git` process is running.

Because our containerized build was running as `root` inside of the
container, git rightfully refused to work on a tree owned by a non-root
user. This commit fixes this issue by creating a release user with the
same UID/GID of the current user (assumed to be the user owning the
working directory), and running `make` with the permissions of that user
instead of running as root.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,10 +25,14 @@ hubble:
 	$(GO_BUILD) $(if $(GO_TAGS),-tags $(GO_TAGS)) -ldflags "-w -s -X 'github.com/cilium/hubble/pkg.GitBranch=${GIT_BRANCH}' -X 'github.com/cilium/hubble/pkg.GitHash=$(GIT_HASH)' -X 'github.com/cilium/hubble/pkg.Version=${VERSION}'" -o $(TARGET)
 
 release:
-	docker run --env "RELEASE_UID=$(RELEASE_UID)" --env "RELEASE_GID=$(RELEASE_GID)" --rm --workdir /hubble --volume `pwd`:/hubble docker.io/library/golang:1.18.3-alpine3.16 \
-		sh -c "apk add --no-cache make git && make local-release"
+	docker run --rm --workdir /hubble --volume `pwd`:/hubble docker.io/library/golang:1.18.3-alpine3.16 \
+		sh -c "apk add --no-cache make git && \
+			addgroup -g $(RELEASE_GID) release && \
+			adduser -u $(RELEASE_UID) -D -G release release && \
+			su release -c 'make local-release'"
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -25,10 +25,14 @@ hubble:
 	$(GO_BUILD) $(if $(GO_TAGS),-tags $(GO_TAGS)) -ldflags "-w -s -X 'github.com/cilium/hubble/pkg.GitBranch=${GIT_BRANCH}' -X 'github.com/cilium/hubble/pkg.GitHash=$(GIT_HASH)' -X 'github.com/cilium/hubble/pkg.Version=${VERSION}'" -o $(TARGET)
 
 release:
-	docker run --env "RELEASE_UID=$(RELEASE_UID)" --env "RELEASE_GID=$(RELEASE_GID)" --rm --workdir /hubble --volume `pwd`:/hubble docker.io/library/golang:1.18.3-alpine3.16 \
-		sh -c "apk add --no-cache make git && make local-release"
+	docker run --rm --workdir /hubble --volume `pwd`:/hubble docker.io/library/golang:1.18.3-alpine3.16 \
+		sh -c "apk add --no-cache make git && \
+			addgroup -g $(RELEASE_GID) release && \
+			adduser -u $(RELEASE_UID) -D -G release release && \
+			su release -c 'make local-release'"
```
tklauser, line 32: FYI, we've had the same issue in cilium-cli and resolved it slightly differently: https://github.com/cilium/cilium-cli/commit/023ccfb8466e2c649230b5cf6c98260bee85c863

Not sure whether that approach is preferable though :thinking:

gandro: Thanks! Yeah, I was aware of `safe.directory`, but since we still need to chmod/chown the built binaries, I figured we might as well switch the user instead.

tklauser: Thanks, that's a good point! We should probably change to the approach taken in this PR for cilium-cli as well.

tklauser (approved):

Makefile
tklauser: FWIW, I've opened cilium/cilium-cli#945 to switch to this PR's approach in cilium-cli as well. Thanks again.

PR #769, opened 2022-07-29
Use command path when registering flagsets

chancez:
Avoids conflicts when different sub-commands have the same command name.
(Eg: hubble status vs hubble foobar status).

Since commands paths rely on having parent-commands, we must initialize
the map containing the flagsets after all sub-commands have been added.

This was something I discovered after trying to add new sub-commands that have the same name as other sub-sub-commands.

### DIFF
--- a/cmd/common/template/usage.go
+++ b/cmd/common/template/usage.go
@@ -17,29 +17,47 @@ package template
 import (
 	"strings"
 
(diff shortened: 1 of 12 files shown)


michi-covalent (approved):
let's also get @rolinh to review it

rolinh (approved):
lgtm

PR #773, opened 2022-08-02
v.09: Update Go to 1.17.13

tklauser:


rolinh:
As per our policy, v0.9 is no longer maintained (see [table](https://github.com/cilium/hubble#releases)). I'd say it's not worth updating Go for this version.

PR #779, opened 2022-08-22
Update Go to 1.19

tklauser:


rolinh:
I think we need to first bump golangci-lint to v1.48.0 for Go 1.19 support.

rolinh (approved):

PR #781, opened 2022-08-22
v0.10: printer: fix argument order in strings.HasPrefix call

tklauser:
[ upstream commit d7910888d26abf817841a1904c5fc6e462f6861f ]

We should check for the "reserved:" prefix in the label string, not vice
versa.

Found by the gocritic linter.


gandro (approved):
Thanks!

Issue #519, opened 2021-03-22
hubble observe doesn't support `--until` without `--since`


michi-covalent:
right now `--until` flag seems to get ignored if `--since` is not specified. i was expecting `--since` to get set to zero if only `--until` flag is specified 🤔 

slack discussion: https://cilium.slack.com/archives/CQRL1EPAA/p1616170452006600

ahmedaabouzied:
Is this still relevant? Can I pick it up?

rolinh:
Yes, this is still relevant. Thanks for volunteering @ahmedaabouzied, I'll assign this issue to you!

chowmean:
Hi, if this is not WIP, can I take it up? @rolinh

kaworu:
@ahmedaabouzied are ou working on this?

ChrsMark:
Hey folks! Not sure what is the status and the assignment of this issue but I took the chance to push a fix for it at https://github.com/cilium/hubble/pull/792. I opened the PR as draft so feel free to close+ignore it if other people are already working on it :).

PR #792, opened 2022-10-01
Fix observe command not supporting --until without --since

ChrsMark:
At the moment, `hubble observe` command ignores `--until` flag if `--since` is not specified. It would be expected `--since` flag to not be a hard requirement for `--until` flag to be used on its own.

This patch makes `--until` flag to be consider without  having  `--since` flag as a prerequisite.

Fixes https://github.com/cilium/hubble/issues/519

Signed-off-by: ChrsMark <chrismarkou92@gmail.com>

### DIFF
--- a/cmd/observe/agent_events.go
+++ b/cmd/observe/agent_events.go
@@ -91,18 +91,18 @@ func getAgentEventsRequest() (*observerpb.GetAgentEventsRequest, error) {
 		if err := since.CheckValid(); err != nil {
 			return nil, fmt.Errorf("failed to convert `since` timestamp to proto: %v", err)
 		}
-		// Set the until field if both --since and --until options are specified and --follow
-		// is not specified. If --since is specified but --until is not, the server sets the
-		// --until option to the current timestamp.
-		if selectorOpts.until != "" && !selectorOpts.follow {
-			ut, err := hubtime.FromString(selectorOpts.until)
-			if err != nil {
-				return nil, fmt.Errorf("failed to parse the until time: %v", err)
-			}
-			until = timestamppb.New(ut)
-			if err := until.CheckValid(); err != nil {
-				return nil, fmt.Errorf("failed to convert `until` timestamp to proto: %v", err)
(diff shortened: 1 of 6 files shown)


cmd/observe/agent_events.go
```diff
@@ -91,18 +91,18 @@ func getAgentEventsRequest() (*observerpb.GetAgentEventsRequest, error) {
...
-			}
-			until = timestamppb.New(ut)
-			if err := until.CheckValid(); err != nil {
-				return nil, fmt.Errorf("failed to convert `until` timestamp to proto: %v", err)
-			}
+	}
+	// Set the until field if both --since and --until options are specified and --follow
+	// is not specified. If --since is specified but --until is not, the server sets the
+	// --until option to the current timestamp.
+	if selectorOpts.until != "" && !selectorOpts.follow {
```
rolinh, line 98: This comment should be update as with the change, `--until` may be provided without `--since`.

cmd/observe/agent_events_test.go
```diff
@@ -67,3 +67,21 @@ func Test_getAgentEventsRequest(t *testing.T) {
 		Until:  timestamppb.New(until),
 	}, req)
 }
+
+func Test_getAgentEventsRequestWithoutSince(t *testing.T) {
```
rolinh, line 71: Thanks for adding tests!

rolinh (commented):
LGTM except for the comment, see below.

cmd/observe/agent_events.go
ChrsMark: Nice catch, thanks! Updated in [9909183](https://github.com/cilium/hubble/pull/792/commits/9909183b871af12a3c64669b6a78693d27aaf539), let me know if the wording is ok or if there is anything to improve :).

rolinh (approved):

kaworu (approved):

michi-covalent (approved):
🚢

Issue #809, opened 2022-11-07
hubble observe does not work with stdin redirection


bimmlerd:
Trying to avoid a [useless use of cat](https://porkmail.org/era/unix/award) I typed `hubble observe < cilium-55s8v-hubble_observe.json`, but was greeted with a cryptic `failed to connect to 'localhost:4245': connection error: desc = "transport: error while dialing: dial tcp [::1]:4245: connect: connection refused"`. Typing `cat cilium-55s8v-hubble_observe.json | hubble observe` worked fine.

I'm guessing that the error stems from it trying and failing to connect to some cluster because tty detection (?) fails in this case. cc @gandro.

bimmlerd:
https://github.com/cilium/hubble/blob/a24aa4b051172ab49f715e94af95c5a60f2c5d2d/cmd/observe/flows.go#L157-L160 looks like the culprit, I suspect input redirection is different from a FIFO pipe?

gandro:
Reproducible on linux-amd64 (archlinux)

rolinh:
I have a patch [here](https://github.com/cilium/hubble/pull/815), could you confirm that it works for you?

bimmlerd:
Works for me, nice 🚀

PR #815, opened 2022-11-14
cmd/observe: fix stdin reading from file redirection

rolinh:
The `hubble observe` command could only read from stdin when being piped to, e.g. `cat sample.json | hubble observe` but not via file redirect such as `hubble observe < sample.json`. This commit makes the latter possible.

Fix: #809


michi-covalent (approved):
neat

PR #816, opened 2022-11-16
Add support for SockLB events 

gandro:
This adds support for [Trace SockLB events](https://github.com/cilium/cilium/pull/21685). These events are similar to L3/L4 trace events, but are traced on the socket level. This means they have a few differences to regular trace events:

 - There are events for pre- and post-translation. This means that we now get visibility into the service load balancing, meaning that source/destination service is populated before NAT/after rev-NAT.
 - These events do not contain the source port or packet related details (such as TCP flags, Ethernet headers etc).
 - Because the events are emitted on a socket level, there is no meaningful traffic direction or reply status. The reply status and traffic direction of the trace sock events is unknown.
 - These events have two verdicts: TRACED and TRANSLATED. The former is used when ever the SockLB BPF hook is executed, the latter is additionally emitted if NAT or reverse NAT has been applied.

Here is an example of `pod-to-a` pod doing a service lookup for the `echo-a` service (first the DNS lookup against kubedns, then a TCP connection to the `echo-a` service):
image


gandro:
> So these are displayed by default right? I wonder: should they be?

I would argue they should. The service IP is not visible anywhere otherwise. Before SockLB was a thing, Hubble would show the clusterIP in `from-endpoint` events (iirc), that visibility got lost till recently.

michi-covalent (approved):

rolinh (approved):
:rocket:

chancez (approved):

gandro:
Somehow GitHub seems to be stuck on the status checks. Closing an re-opening.

kaworu (approved):
Thanks @gandro!

PR #826, opened 2022-12-19
cmd: Make `-o json` an alias for `-o jsobpb`

gandro:
This commit removes the old deprecated JSON output and fully replaces it with the newer JSONPB output (based on proto3's json mapping).

This is a breaking change, users of `-o json` will have to adopt the output parsing slightly: Each emitted object is now wrapped in a `GetFlowsResponse`, which means that the flow object (the root of the previous JSON output) is now accessible via the `.flow` attribute. 

For example, a query of the form `hubble observe -o json | jq .source.labels` now needs to be written as `hubble observe -o json | jq .flow.source.labels`.


kaworu (approved):

christarazi (approved):
It's happening!

PR #836, opened 2023-01-10
dependabot config improvements

kaworu:
See commits, I would also be inclined to bump the `open-pull-requests-limit` (e.g. to 5) and maybe the schedule interval to daily.


rolinh (approved):
Thanks! I also think we could bump the PR limit to 3 or even 5.

tklauser (approved):

PR #849, opened 2023-01-12
github: Enable dependabot for stable branch

gandro:
Hubble CLI currently maintains support for the last stable branch. To ensure security-relevant depencencies are updated, this commit introduces dependabot for the current (v0.11) stable branch with the following configuration:

  - gomod dependencies are only updated if there is a security vulnerability in one of our dependencies.
  - docker dependencies (i.e. the alpine base image)  are only updated to the next patch version
  - github actions are always updated (this mirrors cilium/cilium's configuration)

The goal of this configuration is to ensure we pull in security relevant updates, while keeping the moving parts as low as possible in the stable branch.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,12 +14,57 @@ updates:
     - kind/enhancement
     - release-note/misc
     - priority/release-blocker
+  - package-ecosystem: gomod
+    directory: /
+    schedule:
+      interval: daily
+    commit-message:
+      prefix: "vendor:"
+    # Setting open-pull-requests-limit to 0 means that dependabot will not
+    # update regular dependencies on this target branch, but still provide
+    # security updates for our gomod dependencies
+    open-pull-requests-limit: 0
+    target-branch: "v0.11"
+    rebase-strategy: disabled
+    ignore:
+      - dependency-name: "github.com/cilium/cilium"
+    labels:
+      - kind/enhancement
+      - release-note/misc
+      - priority/release-blocker
+
   - package-ecosystem: github-actions
     directory: /
     schedule:
       interval: daily
     commit-message:
       prefix: "ci:"
+    open-pull-requests-limit: 5
+    rebase-strategy: disabled
+    labels:
+    - kind/enhancement
+    - release-note/misc
+    - priority/release-blocker
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily
+    commit-message:
+      prefix: "ci:"
+    open-pull-requests-limit: 5
+    rebase-strategy: disabled
+    target-branch: "v0.11"
+    labels:
+      - kind/enhancement
+      - release-note/misc
+      - priority/release-blocker
+
+  - package-ecosystem: docker
+    directory: /
+    schedule:
+      interval: daily
+    commit-message:
+      prefix: "dockerfile:"
     open-pull-requests-limit: 5
     rebase-strategy: disabled
     labels:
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -38,3 +71,18 @@ updates:
...
+    directory: /
+    schedule:
+      interval: daily
+    commit-message:
+      prefix: "dockerfile:"
+    open-pull-requests-limit: 5
+    rebase-strategy: disabled
+    target-branch: "v0.10"
+    # Only bump the alpine patch version in stable branches
+    update-types: ["version-update:semver-patch"]
```
kaworu, line 84: [CI is unhappy](https://github.com/cilium/hubble/pull/849/checks?check_run_id=10611332689):

```
The property '#/updates/5/' contains additional properties ["update-types"] outside of the schema when none are allowed
```

According to [the dependabot configuration doc](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#specifying-dependencies-and-versions-to-ignore) `update-types` is only supported in the `ignore:` section, so to achieve the goal of only bumping patch versions I think this should be:

```suggestion
    # Only bump the alpine patch version in stable branches
    ignore:
        update-types: ["version-update:semver-major", "version-update:semver-minor"]
```

kaworu (changes requested):
Thanks @gandro, all the changes make sense to me. Added a comment about `update-types` but other than that LGTM!

.github/dependabot.yml
```diff
@@ -14,6 +14,25 @@ updates:
...
+    directory: /
+    schedule:
+      interval: daily
+    commit-message:
+      prefix: "vendor:"
+    # Setting open-pull-requests-limit to 0 means that dependabot will not
+    # update regular dependencies on this target branch, but still provide
+    # security updates for our gomod dependencies
+    open-pull-requests-limit: 0
+    target-branch: "v0.10"
```
rolinh, line 27:
```suggestion
    target-branch: "v0.11"
```

.github/dependabot.yml
```diff
@@ -26,6 +45,20 @@ updates:
...
     - priority/release-blocker
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily
+    commit-message:
+      prefix: "ci:"
+    open-pull-requests-limit: 5
+    rebase-strategy: disabled
+    target-branch: "v0.10"
```
rolinh, line 56:
```suggestion
    target-branch: "v0.11"
```

.github/dependabot.yml
```diff
@@ -38,3 +71,20 @@ updates:
...
     - priority/release-blocker
+  - package-ecosystem: docker
+    directory: /
+    schedule:
+      interval: daily
+    commit-message:
+      prefix: "dockerfile:"
+    open-pull-requests-limit: 5
+    rebase-strategy: disabled
+    target-branch: "v0.10"
```
rolinh, line 82:
```suggestion
    target-branch: "v0.11"
```

rolinh (changes requested):
Thanks, lgtm except that the current stable branch is v0.11, not v0.10 :slightly_smiling_face:

kaworu (commented):
LGTM besides @rolinh's comment since we released `v0.11`.

kaworu (approved):

.github/dependabot.yml
gandro: :facepalm: Thanks, fixed.

rolinh (approved):

PR #851, opened 2023-01-13
Add hubble observe flows

### LINKED ISSUE #543: Add `hubble observe flows` as alias for `hubble observe`
As we start to introduce additional types of events in Hubble besides flows, we need to distinguish between the different types of events which are being observed. While we want to keep `hubble observe` for compatibility reasons, we canmake it consistent with the newly added debug and agent events by adding a `hubble observe flows` alias:

https://github.com/cilium/hubble/pull/537#issuecomment-830422131

chancez:
Hubble observe is now an alias for `hubble observe flows`. The only major difference is the help text for `hubble observe` is now briefer and indicates it is an alias for `hubble observe flows`. Additionally, while refactoring, I removed the flow formatting flags from the general formatting flagset and just moved it to it's own flagset.

Fixes #543

### DIFF
--- a/cmd/observe/agent_events.go
+++ b/cmd/observe/agent_events.go
@@ -19,14 +19,13 @@ import (
 	"github.com/cilium/hubble/pkg/logger"
 	hubtime "github.com/cilium/hubble/pkg/time"
 	"github.com/spf13/cobra"
-	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
-func newAgentEventsCommand(vp *viper.Viper, flagSets ...*pflag.FlagSet) *cobra.Command {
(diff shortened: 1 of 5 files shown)


kaworu (approved):
Thanks a lot for tackling this @chancez! :pray: Tested and the patch LGTM to me as well.

One funny thing I noticed is that `hubble observe flows` already works before this patch (I guess the missing subcommand invoke the parent because `hubble observe foo` also show flows).

tklauser (approved):
Thanks a lot for picking this up! :rocket:

gandro:
@chancez I think this PR broke the shorthand flags:

```
./hubble observe -o json
unknown shorthand flag: 'o' in -o
```

Only occurs on master, v0.11 is fine.

rolinh:
I think it's shorthand flags, formatting flags seem to be broken:
```
$ hubble observe --output json
unknown flag: --output
$ hubble observe flows --output json
unknown flag: --output
```

chancez:
@rolinh Ah. I guess tests and my manual usage weren't using these flags. Let's revert and I can try again.

PR #865, opened 2023-01-24
cmd: Introduce `HUBBLE_COMPAT=legacy-json-output`

gandro:
This commit introduces a new environment variable, `HUBBLE_COMPAT`, which can be used to restore old behavior in the presence of breaking changes. It's inspired by Golang's GODEBUG variable documented here: https://pkg.go.dev/runtime

The intent of this flag is to allow us introduce breaking changes in the CLI, while still preserving backwards-compatibility with older versions where needed. As an example, this commit also bring back the legacy `-o json` flow schema if `HUBBLE_COMPAT` is set to `legacy-json-output`. This allows older Cilium stable versions to pull in a newer version of the Hubble CLI as part of security updates, while preserving the behavior of the Hubble CLI that originally shipped with that Cilium version.

HUBBLE_COMPAT options are _not_ intended to be supported indefinitely. Rather, compat options may be removed once the ecosystem (such as downstream users like the Cilium-Agent image) has caught up.

We use an environment variable instead of a viper flag to ensure that we could also use it to modify the behavior of viper or pflags if needed in the future.

### DIFF
--- a/cmd/common/config/compat.go
+++ b/cmd/common/config/compat.go
@@ -1,0 +1,44 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Hubble
+
+package config
+
+import (
+	"os"
+	"strings"
+)
+
+const (
+	// HUBBLE_COMPAT is an environment variable similar to GODEBUG.
+	//
+	// It allows us to preserve old CLI behavior in the presence of
+	// breaking changes.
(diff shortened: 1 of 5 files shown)


rolinh (approved):
The approach and implementation look good to me, thanks!

Related to breaking changes, I was wondering whether we shouldn't document somewhere that we don't guarantee backwards compatible changes on the `compact` output or any format other than `json`. What do you think?

PR #869, opened 2023-01-24
Revert "Add hubble observe flows"

chancez:
Reverts cilium/hubble#851

This broke using some flags, so revert until we can fix the regression.


rolinh (approved):

chancez:
Dang that's what I get for clicking the revert button in GH. No sign-off.

PR #873, opened 2023-01-25
.github: Add integration tests that installs cilium and queries it using Hubble CLI

chancez:
I noticed we don't have any tests that actually hit cilium, so this should fix that.

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -1,0 +1,117 @@
+name: Run integration tests
+
+# Any change in triggers needs to be reflected in the concurrency group.
+on:
+  push:
+    branches:
+      - master
+      - v*
+  pull_request:
+    branches:
+    - master
+    - v*
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.event.after }}
+  cancel-in-progress: true
+
+env:
+  HELM_VERSION: v3.10.3
+  KIND_VERSION: v0.17.0
+  KIND_CONFIG: .github/kind-config.yaml
+  CILIUM_VERSION: 1.12.5
+  CILIUM_VALUES: .github/cilium-values.yaml
+
+jobs:
+  integration-test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
+      - name: Setup go
+        uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568
+        with:
+          go-version: '1.19.5'
+      - name: Build hubble CLI
+        run: make
+      - name: Set up Helm
+        uses: azure/setup-helm@5119fcb9089d432beecbf79bb2c7915207344b78
+        with:
+          version: ${{ env.HELM_VERSION }}
+      - name: Create kind cluster
+        uses: helm/kind-action@d8ccf8fb623ce1bb360ae2f45f323d9d5c5e9f00
+        with:
+          version: ${{ env.KIND_VERSION }}
+          config: ${{ env.KIND_CONFIG }}
+          cluster_name: kind
+      - name: Deploy Cilium
+        run: |
+          # Deploy cilium with KPR
+          MASTER_IP="$(docker inspect kind-control-plane | jq '.[0].NetworkSettings.Networks.kind.IPAddress' -r)"
+          helm repo add cilium https://helm.cilium.io
+          helm repo update
+          helm install cilium cilium/cilium \
+            --wait \
+            --namespace kube-system \
+            --version ${{ env.CILIUM_VERSION }} \
+            --values ${{ env.CILIUM_VALUES }} \
+            --set kubeProxyReplacement=strict \
+            --set k8sServiceHost="${MASTER_IP}" \
+            --set k8sServicePort=6443
+      - name: Wait for hubble-relay to be running
+        run: |
+          kubectl -n kube-system rollout status deployment/hubble-relay
+      - name: Run integration tests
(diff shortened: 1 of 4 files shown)


gandro (approved):
Awesome, thanks a lot for doing this!

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,96 @@
...
+        run: |
+          set -ex
+          ./hubble --version
+
+          kubectl -n kube-system port-forward service/hubble-relay 4245:80 &
+          sleep 5s
+          test $(pgrep --count --full "kubectl.*port-forward.*service\/hubble-relay.*4245:80") -eq 1
+
+          # query hubble
+          ./hubble observe -n kube-system
```
gandro, line 75: We can go arbitrarily far with this - I do wonder however if we should check that there is at least some output here, since Hubble will return exit code 0 even if there are no flows.

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,96 @@
...
+        run: |
+          kubectl -n kube-system rollout status deployment/hubble-relay
+      - name: Run integration tests
+        run: |
+          set -ex
+          ./hubble --version
+
+          kubectl -n kube-system port-forward service/hubble-relay 4245:80 &
+          sleep 5s
+          test $(pgrep --count --full "kubectl.*port-forward.*service\/hubble-relay.*4245:80") -eq 1
```
gandro, line 72: Should we use an `until pgrep .. sleep` loop here?

.github/workflows/integration-tests.yaml
kaworu: How about something like

```
./hubble observe -n kube-system -o jsonpb | tee flows.json
test $(jq --slurp length flows.json) -gt 0
```

From there we can re-use `flows.json`, e.g. to test pipe to the CLI

```
test $(./hubble observe < flows.json) -eq $(wc -l flows.json)
```

kaworu (approved):
Thanks @chancez !

.github/workflows/integration-tests.yaml
chancez: Yeah. I copied it from the timescape/rbac CI. It basically never fails, so it's not _necessary_ but might as well.

.github/workflows/integration-tests.yaml
chancez: Yeah I think eventually I'd like to actually move these into Go based tests we run, or even just [bats](https://github.com/bats-core/bats-core/). I can definitely make it check for flows though, that's a good idea.

chancez:
Hmm I can't see to get hubble to return any flows. I wonder if there's something wrong with relay and hubble in the values I'm using.

kaworu:
@chancez pushed a few commits on top of yours, tested locally with `act -j integration-test` and the integration tests succeed on my machine, no idea why it's failing in CI though :sad-parrot:

chancez:
Recent changes, indicate to me that relay isn't connected to the agent:
```
./hubble status
Handling connection for 4245
Current/Max Flows: 0/0
Flows/s: N/A
Connected Nodes: 0/0
```

```
kubectl -n kube-system exec ds/cilium -c cilium-agent -- hubble status
Healthcheck (via unix:///var/run/cilium/hubble.sock): Ok
Current/Max Flows: 85/4,095 (2.08%)
Flows/s: 3.69
```

chancez:
Added more commands to check flows at the post-test info gathering step. It seems if we query hubble status (over port-forward) at the end, the nodes are connected via relay and relay is reporting flows. 

If I then query for flows in the agent, it works, but if I query flows over the port-forward, it doesn't. 
```
Collecting hubble status
+ ./hubble status
Current/Max Flows: 1,593/4,095 (38.90%)
Flows/s: 4.76
Connected Nodes: 1/1
+ kubectl -n kube-system exec ds/cilium -c cilium-agent -- hubble status
Healthcheck (via unix:///var/run/cilium/hubble.sock): Ok
Current/Max Flows: 1,595/4,095 (38.95%)
Flows/s: 4.76
+ echo 'Checking flows 1'
Checking flows 1
+ ./hubble observe -n kube-system -o jsonpb
+ echo 'Checking flows 2'
Checking flows 2
+ kubectl -n kube-system exec ds/cilium -c cilium-agent -- hubble observe -n kube-system -o jsonpb
{"flow": .... .}
```

It's very strange. I thought it was relay, and it might be, but it's able to report back the status, so I'm not suspicious it's just the port-forward, or Hubble CLI itself.

chancez:
Interesting.. with debug logs it indicates it's reading from stdin. I wonder if I broke it (again) with #874
>>>
time="2023-01-27T19:50:25Z" level=debug msg="Reading flows from stdin"
[258](https://github.com/cilium/hubble/actions/runs/4027680636/jobs/6923690090#step:9:259)

chancez:
Okay, I took my work-around from timescape and added it. The key is adding this, which makes the commands running in the step think they're running in a TTY, preventing Hubble CLI from incorrectly assuming no-TTY == data being fed in via stdin.
```
        shell: 'script -q -e -c "bash --noprofile --norc -eo pipefail {0}"'
```

chancez:
It's stuck waiting on `build`, but I think it's because I renamed that action to `unit-test` instead. If anyone can help fix that in the settings @gandro or @rolinh

chancez:
> Not sure I follow. The workflow in question is renamed in _this_ PR. We can ignore the required build in this PR for now and an admin can merge it once it's marked ready-to-merge.

Ah you're right, I was juggling two PRs and thought I renamed it in the other PR. That being said, this PR is ready @gandro

nbusseneau (approved):
LGTM.

gandro:
Thanks! Updated the required pipelines, all open PRs will have to rebase.

PR #874, opened 2023-01-25
Add unit tests for testing hubble args/flags handling

chancez:
Add unit tests that just test the flag/command handling.

This should help prevent what happened in https://github.com/cilium/hubble/pull/851#issuecomment-1402002694.

### DIFF
--- a/cmd/cli_test.go
+++ b/cmd/cli_test.go
@@ -1,0 +1,105 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Hubble
+
+package cmd
+
+import (
+	"bytes"
+	"context"
+	_ "embed"
+	"fmt"
+	"testing"
+
+	"github.com/cilium/cilium/api/v1/observer"
+	"github.com/cilium/hubble/cmd/observe"
+	"github.com/cilium/hubble/pkg/defaults"
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+//go:embed observe_help.txt
+var expectedObserveHelp string
+
+func init() {
+	// Override the client so that it always returns an IOReaderObserver with no flows.
+	observe.GetHubbleClientFunc = func(_ context.Context, _ *viper.Viper) (client observer.ObserverClient, cleanup func() error, err error) {
+		cleanup = func() error { return nil }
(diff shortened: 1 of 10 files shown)


cmd/cli_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	"github.com/cilium/cilium/api/v1/observer"
+	"github.com/cilium/hubble/cmd/observe"
+	"github.com/cilium/hubble/pkg/defaults"
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+var expectedObserveHelp string
```
gandro, line 21: could we use `//go:embed observe_help.txt` here to simplify the code a bit?

gandro (approved):
Thanks, looks good overall! Glad to see coverage increased.

cmd/observe/flows.go
```diff
@@ -136,6 +136,28 @@ func getFlowFiltersYAML(req *observer.GetFlowsRequest) (string, error) {
...
+var GetHubbleClientFunc = func(ctx context.Context, vp *viper.Viper) (client observer.ObserverClient, cleanup func() error, err error) {
+	fi, err := os.Stdin.Stat()
+	if err != nil {
+		return nil, nil, err
+	}
+	if fi.Mode()&os.ModeCharDevice == 0 {
+		// read flows from stdin
+		client = NewIOReaderObserver(os.Stdin)
+		logger.Logger.Debug("Reading flows from stdin")
+	} else {
```
rolinh, line 149: Rather than using an `else` close, I'd return in the `if` clause above. This will make the code easier to read and would allow reviewers to more easily catch that `cleanup` is not setup in `stdin` mode, leading to a panic :wink:

rolinh (changes requested):
Awesome, I definitely welcome more tests!
Lgtm except for the bug that leads to a panic (see comment below).

cmd/observe/flows.go
chancez: Ah good point.

chancez:
Hmm I just realized incidentally that this `init` setting GetHubbleClientFunc will potentially effect all the tests (even outside the package?), so I might revisit that approach first.

chancez:
Actually, I may tackle that after #875 since that refactor will make removing the global a bit easier.

rolinh (approved):

PR #875, opened 2023-01-26
Hubble observe flows

chancez:
Attempt number two for https://github.com/cilium/hubble/pull/851 and https://github.com/cilium/hubble/issues/543. 

Hubble observe is now an alias for hubble observe flows. The only major difference is the help text for hubble observe is now briefer and indicates it is an alias for hubble observe flows. Additionally, while refactoring, I removed the flow formatting flags from the general formatting flagset and just moved it to it's own flagset.

This is currently based on #874 which adds unit tests to ensure things do not break this time.

### DIFF
--- a/cmd/cli_test.go
+++ b/cmd/cli_test.go
@@ -67,7 +67,7 @@ Available Commands:
   config      Modify or view hubble config
   help        Help about any command
   list        List Hubble objects
-  observe     Observe flows of a Hubble server
+  observe     Observe flows and events of a Hubble server
   status      Display status of Hubble server
   version     Display detailed version information
 
--- a/cmd/observe/agent_events.go
+++ b/cmd/observe/agent_events.go
@@ -26,7 +26,7 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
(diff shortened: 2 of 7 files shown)


gandro (commented):
Overall looks fine to me. In local testing, I noticed however that `hubble observe foo` still is an alias for `hubble observe flows`. Is that intended?

chancez:
> Overall looks fine to me. In local testing, I noticed however that `hubble observe foo` still is an alias for `hubble observe flows`. Is that intended?

Hmmm I missed that. Let me take a look

chancez:
I think that's expected, because we don't specify the min/allowed arguments for each command. In the current Hubble CLI you can also do `hubble observe foo`, as well as:
```
hubble list nodes cats dogs any args
```

Because any terminal command can may use these as `args []string` in the `cobra.Command.Run` closures. The reason it works for `hubble observe` is because it's got a `RunE`, and it has sub-commands.

I think we can probably fix this by using `cobra.Command.Args`, but I believe we could do that for all commands in a follow up PR. What do you think?

gandro (approved):
Makes sense. Yeah I think we should do that in a follow-up PR. Thanks!

kaworu (approved):
Thanks! @chancez

chancez:
Lets try this merge thing again 🤞

PR #929, opened 2023-03-13
cmd/observe: improve help message for --first and --all options

rolinh:


cmd/observe/observe.go
```diff
@@ -51,9 +51,9 @@ var (
 )
 
 func init() {
-	selectorFlags.BoolVar(&selectorOpts.all, "all", false, "Get all flows stored in Hubble's buffer")
-	selectorFlags.Uint64Var(&selectorOpts.last, "last", 0, fmt.Sprintf("Get last N flows stored in Hubble's buffer (default %d). When querying against Hubble Relay, this gets N flows per instance of Hubble connected to that Relay", defaults.FlowPrintCount))
-	selectorFlags.Uint64Var(&selectorOpts.first, "first", 0, "Get first N flows stored in Hubble's buffer")
+	selectorFlags.BoolVar(&selectorOpts.all, "all", false, "Get all flows stored in Hubble's buffer. Note: this option may cause Hubble to return a lot of data. It is recommended to only use it along very specific filters to limit the amount of data returned.")
```
kaworu, line 54: nitpicking: IMHO "very specific" is (ironically) vague here; what qualifies as a "very specific filter"? I think it's so env dependent that it's difficult to answer (e.g. a `--since 1m --namespace alex` filter could return a lot of flows, or only a few), and the "warning" note itself is enough.
```suggestion
	selectorFlags.BoolVar(&selectorOpts.all, "all", false, "Get all flows stored in Hubble's buffer. Note: this option may cause Hubble to return a lot of data. It is recommended to only use it along filters to limit the amount of data returned.")
```

kaworu (approved):
Thanks @rolinh

cmd/observe/observe.go
rolinh: Fair point, the wording isn't appropriate. I'll remove the text as suggested.

PR #938, opened 2023-03-14
[v0.11] Prepare for v0.11.3 release

gandro:


rolinh (approved):
Thanks!

Not that it would change anything in this case but should we bump Cilium from v1.13.0-rc5 to v1.13.0?

gandro:
> Thanks!
> 
> Not that it would change anything in this case but should we bump Cilium from v1.13.0-rc5 to v1.13.0?

Good point. But since it's rather cosmetic, let's do that separately for the next release, I wouldn't consider that change a release blocker.

rolinh:
No, indeed (which is why I approved the PR :wink:).

PR #958, opened 2023-03-23
Correctly handle --first/--last when reading flows from a stdin

chancez:
As the title says, when reading flows from stdin, hubble CLI does not correctly handle --first/--last, but this fixes it.

here's a rough demonstration (pardon the jq, it's to condense the JSON output so it's a bit easier to see).
```
(⎈|kind-kind:N/A) ~/p/w/hubble ❯❯❯ ./hubble observe --first 5 < flows.json                                                                                                pr/chancez/file_input_handle_first_last ✭ ◼
Mar 20 16:16:36.501: 10.0.0.139:35044 (host) <- monitoring/prometheus-prometheus-k8s-0:9090 (ID:23751) to-stack FORWARDED (TCP Flags: ACK, PSH)
Mar 20 16:16:36.501: 10.0.0.139:35044 (host) <- monitoring/prometheus-prometheus-k8s-0:9090 (ID:23751) to-stack FORWARDED (TCP Flags: ACK, FIN)
Mar 20 16:16:36.501: 10.0.0.139:35038 (host) <- monitoring/prometheus-prometheus-k8s-0:9090 (ID:23751) to-stack FORWARDED (TCP Flags: ACK, PSH)
Mar 20 16:16:36.501: 10.0.0.139:35038 (host) <- monitoring/prometheus-prometheus-k8s-0:9090 (ID:23751) to-stack FORWARDED (TCP Flags: ACK, FIN)
Mar 20 16:16:36.515: monitoring/prometheus-prometheus-k8s-0:56640 (ID:23751) <- 172.18.0.2:10250 (host) to-endpoint FORWARDED (TCP Flags: ACK)
(⎈|kind-kind:N/A) ~/p/w/hubble ❯❯❯ head -n5 flows.json | jq '"src id: \(.flow.source.identity) dst id: \(.flow.destination.identity)" '                                   pr/chancez/file_input_handle_first_last ✭ ◼
"src id: 23751 dst id: 1"
"src id: 23751 dst id: 1"
"src id: 23751 dst id: 1"
"src id: 23751 dst id: 1"
"src id: 1 dst id: 23751"
(⎈|kind-kind:N/A) ~/p/w/hubble ❯❯❯ ./hubble observe --last 5 < flows.json                                                                                                 pr/chancez/file_input_handle_first_last ✭ ◼
[...]

PR #962, opened 2023-03-27
observe: warn on unknown field while JSON decoding

kaworu:
Before this patch, when parsing JSON flows and encountering an unknown field, the CLI would log a warning and skip the flow.

Unknown fields is not uncommon however, e.g. when using the Hubble CLI against a more recent Hubble server version.

This patch

* emit the flow if successfully parsed while ignoring unknown fields,
* improve the warning message by hinting at upgrading the Hubble CLI,
* make it so the warning message is logged only once.

### DIFF
--- a/cmd/observe/io_reader_observer.go
+++ b/cmd/observe/io_reader_observer.go
@@ -64,10 +64,11 @@ func (o *IOReaderObserver) ServerStatus(_ context.Context, _ *observerpb.ServerS
 type ioReaderClient struct {
 	grpc.ClientStream
 
-	scanner *bufio.Scanner
-	request *observerpb.GetFlowsRequest
-	allow   filters.FilterFuncs
-	deny    filters.FilterFuncs
+	scanner        *bufio.Scanner
+	discardUnknown bool
+	request        *observerpb.GetFlowsRequest
+	allow          filters.FilterFuncs
+	deny           filters.FilterFuncs
 
 	// Used for --last
 	buffer *container.RingBuffer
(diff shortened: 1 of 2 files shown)


kaworu:
Opening as draft as it will conflict with https://github.com/cilium/hubble/pull/958 and also unsure whether logging a warning message is the way to go.

michi-covalent (approved):
time to ship

cmd/observe/io_reader_observer.go
```diff
@@ -59,10 +59,11 @@ func (o *IOReaderObserver) ServerStatus(_ context.Context, _ *observer.ServerSta
 
 // ioReaderClient implements Observer_GetFlowsClient.
 type ioReaderClient struct {
-	scanner *bufio.Scanner
-	request *observer.GetFlowsRequest
-	allow   filters.FilterFuncs
-	deny    filters.FilterFuncs
+	discardUnknown bool
```
glibsm, line 62: how is this going to be set?

cmd/observe/io_reader_observer.go
```diff
@@ -87,7 +88,24 @@ func (c *ioReaderClient) Recv() (*observer.GetFlowsResponse, error) {
...
 		var res observer.GetFlowsResponse
-		err := protojson.Unmarshal(c.scanner.Bytes(), &res)
+		err := protojson.UnmarshalOptions{DiscardUnknown: c.discardUnknown}.Unmarshal(c.scanner.Bytes(), &res)
+		if err != nil && !c.discardUnknown {
+			prevErr := err
+			// the error might be that the JSON data contains an unknown field.
+			// This can happen we attempting to decode flows generated from a
+			// newer Hubble version than the CLI (having introduced a new
+			// field). Retry parsing discarding unknown fields and see whether
+			// the decoding is successful.
```
glibsm, line 98: is there nothing in the error that we can check to make sure it was unknown field, rather than attempting to parse twice?

kaworu: Tracking the error from protobuf it's from https://github.com/cilium/hubble/blob/bb32734273c5f7a90d38dfde0f709867e06518f9/vendor/google.golang.org/protobuf/encoding/protojson/decode.go#L197 and https://github.com/cilium/hubble/blob/bb32734273c5f7a90d38dfde0f709867e06518f9/vendor/google.golang.org/protobuf/encoding/protojson/decode.go#L96-L101

So I see no way to `errors.Is` as it isn't wrapped. Matching on the error message doesn't feel right either.

chancez:
This only seems to update the io reader observer, what about the gRPC observer client?

chancez:
@kaworu Ahhh, I see. Perhaps you could update the PR title/commit message to reflect this is only when ingesting flows from a file/stdin.

chancez (approved):
Code changes LGTM.

PR #976, opened 2023-04-04
observe: traffic direction filter

kaworu:
Before this patch, it was not possible to filter flows by traffic direction.

See also https://github.com/cilium/cilium/pull/24120 for context.


cmd/observe/flows.go
```diff
@@ -418,6 +418,9 @@ func newFlowsCmdHelper(usage cmdUsage, vp *viper.Viper, ofilter *flowFilter) *co
 	filterFlags.Var(filterVar(
 		"to-identity", ofilter,
 		"Show all flows terminating at an endpoint with the given security identity"))
+	filterFlags.Var(filterVar(
+		"traffic-direction", ofilter,
+		"Show all flows in the given traffic direction (either ingress or egress)"))
```
rolinh, line 423: Could we please add completion for the flag's argument given we know it can only be `ingress` or `egress`?

rolinh (approved):

chancez (approved):

sayboras (approved):

PR #977, opened 2023-04-04
SCTP support

kaworu:
Pulling @gandro for the recorder part.


gandro (approved):
Recorder change looks fine to me. Server-side we don't really support it at the moment, but we might in the future (which is why the API already supports it).

rolinh (approved):

PR #996, opened 2023-04-22
Do not fail if unable to provide client certificate when requested

chancez:
When a server requests a client certificate, it's not always required, so do not fail if a client certificate was not configured, instead let the server reject the request instead, in case it supports other authentication methods.


michi-covalent (approved):
😇

kaworu (approved):

rolinh (approved):

michi-covalent:
the button is green. therefore i click.

PR #1002, opened 2023-04-26
Add support for supplying basic authentication credentials

chancez:
Use case: I want to configure my ingress controller to expose hubble relay with basic auth, but Hubble CLI cannot pass basic auth credentials through.

### DIFF
--- a/cmd/common/config/flags.go
+++ b/cmd/common/config/flags.go
@@ -23,6 +23,8 @@ const (
 	KeyTLSClientCertFile = "tls-client-cert-file" // string
 	KeyTLSClientKeyFile  = "tls-client-key-file"  // string
 	KeyTLSServerName     = "tls-server-name"      // string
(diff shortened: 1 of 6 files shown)


cmd/common/config/flags.go
```diff
@@ -23,6 +23,8 @@ const (
 	KeyTLSClientCertFile = "tls-client-cert-file" // string
 	KeyTLSClientKeyFile  = "tls-client-key-file"  // string
 	KeyTLSServerName     = "tls-server-name"      // string
+	KeyBasicAuthUser     = "basic-auth-user"      // string
```
kaworu, line 26: nit: I slightly prefer to spell out username fully, but not feeling too strongly about that one

```suggestion
	KeyBasicAuthUser     = "basic-auth-username"      // string
```

kaworu (approved):

PR #1029, opened 2023-05-11
Fix Renovate datasources

chancez:
Turns out the tags had `v` prefixes that I missed and I didn't notice the dependency dashboard in my fork telling me it had errors.

I've properly tested this in my fork. See my dependency dashboard: https://github.com/chancez/hubble/issues/8 has no errors and my PR here is the same as my master branch of the fork https://github.com/chancez/hubble/commits/master


michi-covalent (approved):
🤖

PR #1041, opened 2023-05-17
Renovate: Ignore pflag

chancez:
See https://github.com/cilium/hubble/pull/1036#issuecomment-1549605905

Closes #1036


rolinh (approved):
Ah good, thanks!

PR #1055, opened 2023-05-26
Remove dependabot configuration in favor of renovate

chancez:
We're using renovate now, good bye dependabot, you did a good job.


glibsm (approved):
👋

PR #1085, opened 2023-06-09
Add golang to matchPackageNames for go deps groups

chancez:
This might fix https://github.com/cilium/hubble/pull/1081


chancez:
Alternatively, I could use `depName=docker.io/library/golang` in the Makefile instead

kaworu (approved):

PR #1086, opened 2023-06-14
Add hubble list namespaces command

chancez:
Now that the Hubble Observer API supports GetNamespaces, implement it in Hubble CLI.

Relates to
- https://github.com/cilium/cilium/issues/25266 
- https://github.com/cilium/cilium/pull/25563

### DIFF
--- a/cmd/list/list.go
+++ b/cmd/list/list.go
@@ -4,11 +4,19 @@
 package list
 
 import (
+	"encoding/json"
+	"fmt"
+	"io"
+
 	"github.com/cilium/hubble/cmd/common/config"
 	"github.com/cilium/hubble/cmd/common/template"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
+
+var listOpts struct {
+	output string
+}
 
 // New creates a new list command.
 func New(vp *viper.Viper) *cobra.Command {
@@ -23,6 +31,16 @@ func New(vp *viper.Viper) *cobra.Command {
 
 	listCmd.AddCommand(
 		newNodeCommand(vp),
+		newNamespacesCommand(vp),
 	)
 	return listCmd
+}
+
+func jsonOutput(buf io.Writer, v interface{}) error {
+	bs, err := json.MarshalIndent(v, "", "  ")
+	if err != nil {
+		return err
+	}
+	_, err = fmt.Fprintln(buf, string(bs))
+	return err
 }
--- a/cmd/list/namespace_test.go
+++ b/cmd/list/namespace_test.go
@@ -1,0 +1,124 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Hubble
+
+package list
+
+import (
+	"bytes"
+	"testing"
+
+	observerpb "github.com/cilium/cilium/api/v1/observer"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNamespaceOutputs(t *testing.T) {
+	var testCases = []struct {
+		name                    string
+		namespaces              []*observerpb.Namespace
+		expectedJSONOutput      string
+		expectedTableOutput     string
+		expectedWideTableOutput string
+	}{
+		{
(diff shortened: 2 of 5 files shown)


chancez:
> static check is acting starnge. It doesn't appear to pick up your function rename 🤔

Oh I forgot to update the tests after changing the name of a function.

sayboras (approved):
LGTM :heavy_check_mark:

The release note label is major, but I leave it to Hubble team to decide :thinking:

tklauser (approved):

cmd/observe/io_reader_observer.go
```diff
@@ -55,6 +55,11 @@ func (o *IOReaderObserver) GetNodes(_ context.Context, _ *observerpb.GetNodesReq
 	return nil, status.Errorf(codes.Unimplemented, "GetNodes not implemented")
 }
 
+// GetNamespaces is not implemented, and will throw an error if used.
```
kaworu, line 58: "throw" :smile:

kaworu (approved):
Thanks @chancez! Tested the patch and LGTM.

One thing I noticed that bugged me:
```
% ./hubble list namespaces -o wide
NAMESPACE            CLUSTER
kube-system          N/A
local-path-storage   N/A
```

Here my cluster name is "default", but I see N/A. I wonder if we should consider removing [this guard](https://github.com/cilium/cilium/blob/00d8462de1b536195158aee1e67d53c19c18363a/pkg/node/types/nodename.go#L49) and consistently output "default" instead of empty string, although on the other hand it will go against [this CFP](https://github.com/cilium/cilium/issues/23111).

cmd/list/list.go
```diff
@@ -23,6 +31,16 @@ func New(vp *viper.Viper) *cobra.Command {
...
 	)
 	return listCmd
 }
+
+func jsonOutput(buf io.Writer, v interface{}) error {
+	bs, err := json.MarshalIndent(v, "", "  ")
+	if err != nil {
+		return err
+	}
+	_, err = fmt.Fprintln(buf, string(bs))
```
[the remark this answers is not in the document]
chancez: This was how it was already written, I just moved which file this function was in. I believe the goal is to get a newline at the end so that when it's logged it's not on the same line as the prompt.

cmd/observe/io_reader_observer.go
chancez: I copied the existing comments 😂

kaworu:
> I wasn't sure if the client should be doing that or not.

Can the client disambiguate between `""` and `"default"`?

> I kinda feel like perhaps I should be extracting the cluster name from the flows inside hubble server, and configuring the cluster name there. What do you think?

I don't understand what you are suggesting, could you clarify?

chancez:
> Can the client disambiguate between "" and "default"?

We don't really define anywhere how they should be treated, so it's hard to say. 

In the [NodeNameFilter](https://github.com/cilium/cilium/blob/766e62b83c20789fe2ecf35d4766eb27bf1782f7/pkg/hubble/filters/nodename.go#L28-L31) we basically treat empty as "default" for filtering purposes. But in the node_name we field we return in `GetFlows` doesn't get updated from "" to "default" IIRC. So basically, in the server, we always leave the flow alone, but when querying, we support treating "default" as "". 

After writing that out, it makes me think the best option then is to just return "" instead of "N/A". What do you think?

> I don't understand what you are suggesting, could you clarify?

Right now, the clusterName for a flow is determined by using [nodeTypes.GetClusterName()](https://github.com/cilium/cilium/blob/766e62b83c20789fe2ecf35d4766eb27bf1782f7/pkg/hubble/observer/local_observer.go#L630)  which uses [configuration](https://github.com/cilium/cilium/blob/766e62b83c20789fe2ecf35d4766eb27bf1782f7/pkg/node/types/nodename.go#L47) to determine the cluster_name.

I was thinking that we could use the value from the flow itself (which is part of the node_name field), but I realize now it's not much better.

chancez:
Rebased and removed `N/A` in favor of leaving cluster empty if not set.

kaworu (approved):

PR #1101, opened 2023-06-21
Add experimental-field-mask flags

AwesomePatrol:
Follow-up to change in cilium exposing a new experimental field in `GetFlowsRequest`s: https://github.com/cilium/cilium/pull/23198

Usage example with custom field mask and JSON output format:
```
% hubble observe --experimental-field-mask time,verdict --output jsonpb
```

Usage example with default field mask and default output format:
```
% hubble observe --experimental-use-default-field-masks
```

Example of providing invalid fieldmask:
```
% hubble observe --experimental-field-mask time,verdict,invalid --output jsonpb
failed to construct field mask: proto: invalid path "invalid" for message "flow.Flow"
```

Example of providing custom field mask for non-JSON output format:
```
% hubble observe --experimental-field-mask time,verdict
compact output format is not compatible with custom field mask
```

### DIFF
--- a/cmd/observe/flows.go
+++ b/cmd/observe/flows.go
@@ -550,6 +550,7 @@ func handleFlowArgs(writer io.Writer, ofilter *flowFilter, debug bool) (err erro
 		hubprinter.WithColor(formattingOpts.color),
 	}
 
+	jsonOut := false
 	switch formattingOpts.output {
 	case "compact":
 		opts = append(opts, hubprinter.Compact())
@@ -563,6 +564,7 @@ func handleFlowArgs(writer io.Writer, ofilter *flowFilter, debug bool) (err erro
 		fallthrough
 	case "jsonpb":
 		opts = append(opts, hubprinter.JSONPB())
+		jsonOut = true
 	case "tab", "table":
 		if selectorOpts.follow {
 			return fmt.Errorf("table output format is not compatible with follow mode")
@@ -571,6 +573,15 @@ func handleFlowArgs(writer io.Writer, ofilter *flowFilter, debug bool) (err erro
 	default:
 		return fmt.Errorf("invalid output format: %s", formattingOpts.output)
 	}
+	if !jsonOut {
+		if len(experimentalOpts.fieldMask) > 0 {
+			return fmt.Errorf("%s output format is not compatible with custom field mask", formattingOpts.output)
+		}
+		if experimentalOpts.useDefaultMasks {
+			experimentalOpts.fieldMask = defaults.FieldMask
+		}
+	}
+
 	if otherOpts.ignoreStderr {
 		opts = append(opts, hubprinter.IgnoreStderr())
 	}
--- a/cmd/observe/flows.go
+++ b/cmd/observe/flows.go
@@ -30,6 +30,7 @@ import (
 	"github.com/spf13/viper"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 	"gopkg.in/yaml.v3"
 )
@@ -699,6 +700,16 @@ func getFlowsRequest(ofilter *flowFilter, allowlist []string, denylist []string)
 		Since:     since,
 		Until:     until,
 		First:     first,
+	}
+
+	if len(experimentalOpts.fieldMask) > 0 {
+		fm, err := fieldmaskpb.New(&flowpb.Flow{}, experimentalOpts.fieldMask...)
+		if err != nil {
+			return nil, fmt.Errorf("failed to construct field mask: %w", err)
+		}
+		req.Experimental = &observerpb.GetFlowsRequest_Experimental{
+			FieldMask: fm,
+		}
 	}
 
 	return req, nil
--- a/cmd/observe/observe.go
+++ b/cmd/observe/observe.go
@@ -43,7 +43,8 @@ var (
 	}
 
 	experimentalOpts struct {
-		fieldMask []string
+		fieldMask       []string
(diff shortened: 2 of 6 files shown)


cmd/observe/observe.go
```diff
@@ -37,9 +37,10 @@ var (
 	}
 
 	otherOpts struct {
-		ignoreStderr    bool
-		printRawFilters bool
-		inputFile       string
+		ignoreStderr          bool
+		printRawFilters       bool
+		inputFile             string
+		experimentalFieldMask string
```
glibsm, line 43: nit: making a new struct here for experimental opts is probably more clear in the long run

glibsm (approved):

cmd/observe/observe.go
```diff
@@ -150,6 +154,9 @@ func init() {
 
 	otherFlags.StringVar(&otherOpts.inputFile, "input-file", "",
 		"Query flows from this file instead of the server. Use '-' to read from stdin.")
+
+	otherFlags.StringVar(&experimentalOpts.fieldMask, "experimental-field-mask", "",
```
chancez, line 158: This should be a `StringSliceVar`.

chancez (approved):

kaworu (approved):

kaworu:
@chancez I'm hesitant to tag an experimental feature as `release-note/major`, so switching to `release-note/minor` instead. It won't have effect until the next release anyway, so please reply if you feel strongly about this.

kaworu (changes requested):
[CI is unhappy](https://github.com/cilium/hubble/actions/runs/5342685079/jobs/9712330003?pr=1101#step:6:26) about the observe help output test, please update [cmd/observe_help.txt](https://github.com/cilium/hubble/blob/4daf0d0d340da47c2e8e8a1db6d9cbe29cfa4c74/cmd/observe_help.txt) accordingly.

```
go run ./main.go observe --help > cmd/observe_help.txt
git add -p cmd/observe_help.txt
```
(avoid staging the hunk about the `--config` flag, keeping the `%s` template).

kaworu (changes requested):
Thanks for the test fix @AwesomePatrol, one more thing:

Using the experimental flag (e.g. `hubble observe --experimental-field-mask verdict`) works well in combination with `--output json`, less so with other output formats:

```
% ./hubble observe --last 1 --experimental-field-mask verdict --output compact
N/A:  (unknown) <>  (unknown) UNKNOWN FORWARDED ()
N/A:  (unknown) <>  (unknown) UNKNOWN FORWARDED ()
% ./hubble observe --last 1 --experimental-field-mask verdict --output dict
  TIMESTAMP: N/A
     SOURCE:
DESTINATION:
       TYPE: UNKNOWN
    VERDICT: FORWARDED
    SUMMARY:
------------
  TIMESTAMP: N/A
     SOURCE:
DESTINATION:
       TYPE: UNKNOWN
    VERDICT: FORWARDED
    SUMMARY:
% ./hubble observe --last 1 --experimental-field-mask verdict --output tab
TIMESTAMP   SOURCE   DESTINATION   TYPE      VERDICT     SUMMARY
N/A                                UNKNOWN   FORWARDED
N/A                                UNKNOWN   FORWARDED
% ./hubble observe --last 1 --experimental-field-mask verdict --output json
{"flow":{"verdict":"FORWARDED"}}
{"flow":{"verdict":"FORWARDED"}}
```

I think we should
1. disallow `--experimental-field-mask` when the output format is not JSON.
2. For compact and dict/tab consider hardcoding a mask requesting only the fields used by the output format (see [WriteProtoFlow](https://github.com/cilium/hubble/blob/02b21ca708434e56b75941e96a26d0e69b94a83d/pkg/printer/printer.go#L295-L296)).

AwesomePatrol:
I added another flag `--experimental-use-default-field-masks` that specifies if we should use the new experimental field. It is disabled by default. Given backwards compatibility from protobuf it is likely safe to default to true.

I also added a test to verify that all fields needed by output formats are part of the default field mask (which currently captures a bit more data than necessary).

cmd/observe/flows.go
```diff
@@ -570,6 +573,15 @@ func handleFlowArgs(writer io.Writer, ofilter *flowFilter, debug bool) (err erro
 	default:
 		return fmt.Errorf("invalid output format: %s", formattingOpts.output)
 	}
+	if !jsonOut {
+		if len(experimentalOpts.fieldMask) > 0 {
+			return fmt.Errorf("%s output format is not compatible with custom field mask", formattingOpts.output)
+		}
+		if len(experimentalOpts.fieldMask) == 0 && experimentalOpts.useDefaultMasks {
```
kaworu, line 580: nit: because of the previous check, the first condition can never be true
```suggestion
		if experimentalOpts.useDefaultMasks {
```

cmd/observe/observe.go
```diff
@@ -150,6 +155,12 @@ func init() {
 
 	otherFlags.StringVar(&otherOpts.inputFile, "input-file", "",
 		"Query flows from this file instead of the server. Use '-' to read from stdin.")
+
+	otherFlags.StringSliceVar(&experimentalOpts.fieldMask, "experimental-field-mask", nil,
+		"Experimental: Comma-separated list of fields for mask. Fields not in the mask will be removed from server response.")
+
+	otherFlags.BoolVar(&experimentalOpts.useDefaultMasks, "experimental-use-default-field-masks", false,
+		"Experimental: Use field mask to request from server only information that will be displayed")
```
kaworu, line 163: nit
```suggestion
		"Experimental: request only visible fields when the output format is compact, tab, or dict.")
```

kaworu (changes requested):
Thanks for the quick update @AwesomePatrol :tada: 

Couple of nitpicking comments, other than that LGTM.

kaworu (approved):

PR #1108, opened 2023-07-04
treewide: switch the logger to slog

rolinh:
Go v1.21 structured logging is provided by the standard library via the `log/slog` package. Earlier Go releases can use `golang.org/x/exp/slog` instead. Go v1.21 should be released in a few days. At this point, we will be able to get rid of `golang.org/x/exp/slog` as a dependency.

This patch updates the global logger to use slog instead of logrus. For the end users, there are minor differences in the log output. Below is an example of logs as formatted prior to this patch:

    time="2023-07-04T15:59:10+02:00" level=debug msg="Using config file" config-file=/home/bob/.config/hubble/config.yaml
    time="2023-07-04T15:59:10+02:00" level=debug msg="connected to Hubble API" server=server
    time="2023-07-04T15:59:10+02:00" level=debug msg="Sending GetFlows request" request="number:20"

With this patch applied, the is as below:

    time=2023-07-04T15:59:36.472+02:00 level=DEBUG msg="Using config file" config-file=/home/bob/.config/hubble/config.yaml
    time=2023-07-04T15:59:36.924+02:00 level=DEBUG msg="connected to Hubble API" server=server
    time=2023-07-04T15:59:36.924+02:00 level=DEBUG msg="Sending GetFlows request" request=number:20

As you can see, levels are now capitalized. Moreover, the time now is in millisecond resolution rather than second. I personally think that millisecond resolution is a better default. As for the other key/value pairs, they should remain absolutely identical.


lambdanis (approved):

kaworu (approved):

aanm (approved):

rolinh:
I removed the blocked label as Hubble v0.12.0 has now been released. I also converted it to draft as I want to take a chance to test the new slog analyzer that I've implemented (see https://github.com/cilium/linters/pull/2).

michi-covalent (approved):

rolinh:
Scratch that, I'll do it as a follow-up.

PR #1153, opened 2023-07-26
Don't set --last if --input-file is specified

michi-covalent:
Don't set --last to 20 if --input-file flag is specified to preserve the previous default behavior of processing all the flows from stdin.

### DIFF
--- a/cmd/observe/flows_test.go
+++ b/cmd/observe/flows_test.go
@@ -174,3 +174,18 @@ func Test_getFlowsRequest_ExperimentalFieldMask_non_json_output(t *testing.T) {
 	err := handleFlowArgs(os.Stdout, filter, false)
 	assert.ErrorContains(t, err, "not compatible")
 }
+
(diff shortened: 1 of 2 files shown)


michi-covalent:
looking into the test failure 😶‍🌫️

cmd/observe/flows_test.go
```diff
@@ -174,3 +174,12 @@ func Test_getFlowsRequest_ExperimentalFieldMask_non_json_output(t *testing.T) {
 	err := handleFlowArgs(os.Stdout, filter, false)
 	assert.ErrorContains(t, err, "not compatible")
 }
+
+// Don't set the number flag in GetFlowsRequest if --input-file is specified
+func Test_getFlowsRequestWithInputFile(t *testing.T) {
```
chancez, line 179: Can we have a test where --last is used with --input-file too?

michi-covalent: good idea 🥰

chancez (approved):

PR #1182, opened 2023-08-09
treewide: use log/slog instead of golang.org/x/exp/slog

rolinh:
Now that we are using Go v1.21, we can use the `log/slog` from the standard library instead of the experimental package.

This is a follow-up to #1108


rolinh:
The static check failure is due to golangci-lint being built with Go 1.20 and not Go 1.21 meaning it does not understand the new built-ins (`min` and `max` notably) that are introduced with this new version and used in the standard library.
I have locally rebuilt golangci-lint with Go v1.21.0 and can confirm that it doesn't complain after that. So I think we have a couple of options:

- Wait for a golangci-lint release (I'd expect one to happen today or tomorrow)
- Revert the PR that updated Go to v1.21.0

tklauser:
> The static check failure is due to golangci-lint being built with Go 1.20 and not Go 1.21 meaning it does not understand the new built-ins (`min` and `max` notably) that are introduced with this new version and used in the standard library. I have locally rebuilt golangci-lint with Go v1.21.0 and can confirm that it doesn't complain after that. So I think we have a couple of options:
> 
> * Wait for a golangci-lint release (I'd expect one to happen today or tomorrow)

I'd prefer this option, given a golangci-lint release will happen later this week. We can still revert the Go 1.21.0 update if the golangci-lint release is delayed for some reason.

rolinh:
Nice, v1.54.0 was just released: https://github.com/golangci/golangci-lint/releases/tag/v1.54.0. Let's trigger renovate to pick up the update.

tklauser (approved):
Neat, thanks!

Issue #1201, opened 2023-09-05
Raw-Filters Flags are silently dropped for `hubble observe`


glrf:
`hubble observe` silently drops the `--denylist` and `--allowlist` flags. These flags only work for `hubble observe flows`, even though `hubble observe` should just be an alias for `hubble observe flows`.

PR #1202, opened 2023-09-05
Fix raw-filter flags by binding flags to viper only if command is run

glrf:
The raw-filter flags are broken since we bind them to viper twice with the same name, during initalization. Once for `hubble observe` and once for `hubble observe flows`. This means that only the flags for `hubble observe flows` work, as it is registered later and overwrites the binding for `hubble observe`

We fix this by only binding the flags of the command we actually run.

Fixes: #1201

### DIFF
--- a/cmd/cli_test.go
+++ b/cmd/cli_test.go
@@ -31,6 +31,13 @@ func init() {
 	expectedObserveHelp = fmt.Sprintf(expectedObserveHelp, defaults.ConfigFile)
 }
 
+var observeRawFilterArgs = []string{"--allowlist", `{"source_pod":["kube-system/"]}`, "--denylist", `{"source_ip":["1.1.1.1"]}`, "--print-raw-filters"}
+var observeRawFilterOut = `allowlist:
+    - '{"source_pod":["kube-system/"]}'
+denylist:
+    - '{"source_ip":["1.1.1.1"]}'
+`
+
 func TestTestHubbleObserve(t *testing.T) {
 	tests := []struct {
 		name           string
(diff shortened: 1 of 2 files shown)


glrf:
I'm not entirely sure this is the best way to fix this. It's probably the shortest, but it doesn't remove the original footgun that we can easily bind two flagsets to the same viper config.
Maybe someone has a better idea?

chancez:
This approach seems reasonable. I'm sure there's probably a "better" way to do it, but I suspect it involves bigger refactors. 

I do think what we need however is a test for this. Could you add a new test case into https://github.com/cilium/hubble/blob/main/cmd/cli_test.go where you call `hubble observe --allowlist ... --print-raw-filters` and verify the output contains the expected "raw filters"? I think that's probably the simplest. We should have a test case for both `hubble observe` and `hubble observe flows` to make sure they're both getting the raw filters.

glrf:
> I do think what we need however is a test for this. Could you add a new test case into https://github.com/cilium/hubble/blob/main/cmd/cli_test.go where you call `hubble observe --allowlist ... --print-raw-filters` and verify the output contains the expected "raw filters"? I think that's probably the simplest. We should have a test case for both `hubble observe` and `hubble observe flows` to make sure they're both getting the raw filters.

Good point, I totally forgot about that. I added a test case for it. Had to do a very minor change to printing raw filters toactually use the output defined by the command, so that the test case actually works.

chancez (approved):

kaworu (approved):
Nice catch!

PR #1203, opened 2023-09-06
Add filter flag documentation and examples to help message

glrf:
The Hubble CLI provides multiple different flags to filter the returned
flows. However how these filters are combined is not entirely obvious.

This commit adds more detailed help message and multiple examples that
explain how these filters interact if more than one filter flag is
provided.

I struggled to make this concise and understandable, so I added more detailed examples. I'm happy about any input to improve this and make it clearer.

### DIFF
--- a/cmd/observe/flows.go
+++ b/cmd/observe/flows.go
@@ -204,13 +204,77 @@ func newFlowsCmdWithFilter(vp *viper.Viper, ofilter *flowFilter) *cobra.Command 
 
   Note that the observe command ignores --follow, --last, and server flags when it
   reads flows from stdin. The observe command processes and output flows in the same
-  order they are read from stdin without sorting them by timestamp.`,
+  order they are read from stdin without sorting them by timestamp.
+
+* Filtering flows
+
+  Observe provides a long list of filter options. These options let you, for example,
+  filter for the used HTTP method using the '--http-method' flag. The following
+  command shows all flows that use the HTTP PUT method.
+
+    hubble observe --http-method put
+
+  You can also provide multiple values for a flag, in which case a flow matches the
+  filter if it matches any of the provided values. If you add a second '--http-method'
+  flag matching GET requests to the previous example, the command will show any flow
+  using either a PUT or GET method.
+
+    hubble observe --http-method put --http-method get
+
+  If you add a different flag, a flow is only returned if it matches for both of the
+  different flags. For example, you can add a '--to-namespace' flag to the previous
+  command so only flows using GET or PUT requests to endpoints in namespace 'foo' are
+  returned.
+
+    hubble observe --http-method put --http-method get --to-namespace foo
+
+  And by adding another '--to-namespace' flag, it will return flows using GET or PUT
+  requests to endpoints in namespace 'foo' or 'bar'
+
+    hubble observe --http-method put --http-method get --to-namespace foo --to-namespace bar
+
+* Using negations on filters
+
+  Observe can also return all flows that don't match a certain filter by using the
+  '--not' flag. The following command returns all flows that don't use the HTTP PUT
+  method.
+
+    hubble observe --not --http-method put
+
+  To filter out multiple values, you can combine negative flags by prefixing each of
+  them with '--not'. The following command will return all flows that neither use a
+  GET nor PUT method.
+
+    hubble observe --not --http-method put --not --http-method get
+
+  You can also filter using multiple different negative flags. The following example
+  filters out all flows that match both of the flags. For example, by adding a
+  '--not --to-namespace foo' flag, the example command will show all flows that
+  don't use a GET or PUT method to endpoints in the 'foo' namespace.
+
+    hubble observe --not --http-method put --not --http-method get --not --to-namespace foo
+
+  This means the command will still return flows to the 'foo' namespace that don't
+  use HTTP PUT or GET methods, and it will return flows using HTTP PUT and GET
+  methods that end in other namespaces.
+  `,
 		use:   "flows",
 		short: "Observe flows of a Hubble server",
 		long: `Observe provides visibility into flow information on the network and
 application level. Rich filtering enable observing specific flows related to
 individual pods, services, TCP connections, DNS queries, HTTP requests and
-more.`,
+more.
+
+Observe can filter flows using multiple different flags, such as filtering for
+the source namespace or the protocol used. To match a filter flag, a flow must
+match at least one of the provided values for that flag. A flow must match all
+the provided filter flags to be returned.
+
+Observe can also show all flows that do not match a provided filter, by adding
+the '--not' flag in front of a filter flag. To match a negated filter
+flag, a flow must not match any of the provided values for that flag. A returned
+flow does not match any of the negated filter flags.
+`,
 	}
 
 	return newFlowsCmdHelper(usage, vp, ofilter)
--- a/cmd/observe/flows.go
+++ b/cmd/observe/flows.go
@@ -200,7 +200,7 @@ func newFlowsCmdWithFilter(vp *viper.Viper, ofilter *flowFilter) *cobra.Command 
 
   Then,
 
-    cat flows.json | hubble observe
+    cat flows.json | hubble observe --input-file -
 
   Note that the observe command ignores --follow, --last, and server flags when it
   reads flows from stdin. The observe command processes and output flows in the same


cmd/observe/flows.go
```diff
@@ -200,17 +200,80 @@ func newFlowsCmdWithFilter(vp *viper.Viper, ofilter *flowFilter) *cobra.Command
 
   Then,
 
-    cat flows.json | hubble observe
+    cat flows.json | hubble observe --input-file -
```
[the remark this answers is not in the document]
chancez: This is fine. We recently changed this, so the example was forgotten.

cmd/observe/flows.go
```diff
@@ -200,17 +200,79 @@ func newFlowsCmdWithFilter(vp *viper.Viper, ofilter *flowFilter) *cobra.Command
...
 
   Note that the observe command ignores --follow, --last, and server flags when it
   reads flows from stdin. The observe command processes and output flows in the same
-  order they are read from stdin without sorting them by timestamp.`,
+  order they are read from stdin without sorting them by timestamp.
+
+* Filtering flows
+
+  Observe provide a long list of filter options. These option let us, for example,
```
chancez, line 211: Fix grammar, prefer `you` over `we`.

```suggestion
  Observe provides a long list of filter options. These options let you, for example,
```

cmd/observe/flows.go
```diff
@@ -200,17 +200,79 @@ func newFlowsCmdWithFilter(vp *viper.Viper, ofilter *flowFilter) *cobra.Command
...
+  filter for the used HTTP method using the '--http-method' flag. The following
+  command shows all flows that use the HTTP PUT method.
+
+    hubble observe --http-method put
+
+  You can also provide multiple values for a flag, in which case a flow matches the
+  filter if it matches any of the provided values. For our example if we add a second
```
chancez, line 218: we -> you
```suggestion
  filter if it matches any of the provided values. For our example if you add a second
```

cmd/observe/flows.go
```diff
@@ -200,17 +200,79 @@ func newFlowsCmdWithFilter(vp *viper.Viper, ofilter *flowFilter) *cobra.Command
...
+  '--http-method' flag matching GET requests, the command will show any flow that
+  uses either a PUT or GET method.
+
+    hubble observe --http-method put --http-method get
+
+  If we add a different flag, a flow is only returned if it matches for both of the
+  different flags. For example we can add a '--to-namespace' flag to our command so
+  only flows using GET or PUT requests to endpoint in namespace 'foo' are returned.
```
chancez, line 226: we -> you
```suggestion
  If you add a different flag, a flow is only returned if it matches for both of the
  different flags. For example you can add a '--to-namespace' flag to our command so
  only flows using GET or PUT requests to endpoint in namespace 'foo' are returned.
```

cmd/observe/flows.go
```diff
@@ -200,17 +200,79 @@ func newFlowsCmdWithFilter(vp *viper.Viper, ofilter *flowFilter) *cobra.Command
...
+
+    hubble observe --http-method put --http-method get --to-namespace foo
+
+  And by adding another '--to-namespace' flag, it will return flows using GET or PUT
+  requests to endpoint in namespace 'foo' or 'bar'
+
+    hubble observe --http-method put --http-method get --to-namespace foo --to-namespace bar
+
+* Using reversed filters
```
chancez, line 235:
```suggestion
* Using negations on filters
```

cmd/observe/flows.go
```diff
@@ -200,17 +200,79 @@ func newFlowsCmdWithFilter(vp *viper.Viper, ofilter *flowFilter) *cobra.Command
...
+
+  Observe can also return all flows that don't match a certain filter by using the
+  '--not' flag. The following command returns all flows that don't use the HTTP PUT
+  method.
+
+    hubble observe --not --http-method put
+
+  Similarly to filtering for multiple values, we can also filter our multiple values.
```
chancez, line 243: we -> you
```suggestion
  Similarly to filtering for multiple values, you can also filter our multiple values.
```

cmd/observe/flows.go
```diff
@@ -200,17 +200,79 @@ func newFlowsCmdWithFilter(vp *viper.Viper, ofilter *flowFilter) *cobra.Command
...
+  The following command will return all flows that neither use a GET nor PUT method.
+
+    hubble observe --not --http-method put --not --http-method get
+
+  We can also filter using multiple reversed flags. This will then filter out all
```
chancez, line 248: We -> You
```suggestion
 You can also filter using multiple reversed flags. This will then filter out all
```

cmd/observe/flows.go
```diff
@@ -200,17 +200,79 @@ func newFlowsCmdWithFilter(vp *viper.Viper, ofilter *flowFilter) *cobra.Command
...
 		use:   "flows",
 		short: "Observe flows of a Hubble server",
 		long: `Observe provides visibility into flow information on the network and
 application level. Rich filtering enable observing specific flows related to
 individual pods, services, TCP connections, DNS queries, HTTP requests and
-more.`,
+more.
+
+Observe can filter flows using multiple different flags, such as filtering for
+the source namespace or the used protocol. To match a filter flag, a flow most
```
chancez, line 267:
```suggestion
the source namespace or the protocol used. To match a filter flag, a flow most
```

cmd/observe/flows.go
```diff
@@ -200,17 +200,79 @@ func newFlowsCmdWithFilter(vp *viper.Viper, ofilter *flowFilter) *cobra.Command
...
+match at least one of the provided values for that flag. A flow must match all
+the provided filter flags to be returned.
+
+Observe can also show all flows that do not match a provided filter, by adding
+the '--not' flag in front of a filter flag. To not match a reversed filter
```
chancez, line 272: Reversed -> negated and remove double negative.
```suggestion
the '--not' flag in front of a filter flag. To match a negated filter
```

cmd/observe/flows.go
```diff
@@ -200,17 +200,79 @@ func newFlowsCmdWithFilter(vp *viper.Viper, ofilter *flowFilter) *cobra.Command
...
+flag, a flow must not match any of the provided values for that flag. A returned
+flow does not match at least one of the reversed filter flags.
```
chancez, line 274:
```suggestion
flow does not match any of the negated filter flags.
```

glrf:
> My only concern is this is making the already really long `--help` even longer, but I think this is fine.

That's a fair point. The help message is very long now. If this grows even more, it might be a good idea to move the examples to the docs.cilium.io and just link to them in the help message?

chancez:
@gandro @kaworu any thoughts on if we're okay with adding some more usage examples to the help test?

gandro:
> @gandro @kaworu any thoughts on if we're okay with adding some more usage examples to the help test?

Yeah, I'm also a bit worried about help page length. On the other hand, I do think those examples brings a lot of value to users who are calling `hubble observe --help` for the first time. And I think having it displayed to users when they are actively looking for help is a nice way to discover it too (since we have to assume that many people will not read or find the docs as quickly as they will try `--help`).

It's mostly users who already know the basics of Hubble for which this adds a lot of visual clutter, making it harder to quickly look up the reference for a flag they are looking for.

Maybe an option would be to maybe have short and long help flag? That I think would cover both cases: Users who are learning Hubble and want a more verbose help page, vs. users who just want to quickly know about the syntax or name of a particular flag.

kaworu (changes requested):
Thanks @glrf! Couple of typos and nits on the way.

Overall LGTM, agree with @gandro that a short and long help would be best.

glrf:
Thanks for all the input on the help text. I often struggle with technical writing so I'm always happy about feedback.

And about the short and long help flag. I think I agree that this sounds smart, but when I actually tried to implement this I struggled to find a clean solution. I can see three ways of doing this and all three have pros and cons

* Make `-h,--help` the short help flag and introduce something like `--long-help` or `--detailed-help` or `--explain` that shows the long help text. The disadvantage here would be that new users might just not see it. Similar problem to moving it to the docs and linking.
* Make `-h,--help` the long help flag and introduce something like `-b,--brief-help` or mabye `--ref`(for reference? Or IDK naming is hard). The disadvantage here is that most people will likely just continue using `-h` and live with the clutter, as they're just used to it.
* Make `-h` show the short help text and `--help` show the long help text. I could make this work, but it felt like a hack and I don't know if that's too magical and will confuse users.

Any opinions?

chancez:
Another alternative is a new sub-command like `hubble observe usage-examples` or something?

gandro (approved):

kaworu (approved):

chancez (approved):

PR #1244, opened 2023-10-10
renovate: Use allowedVersions to limit updates to stable branches

chancez:
`matchUpdateTypes` is the wrong option for limiting updates in the stable branch. When I wrote this config, it was based off Cilium's config, which I looked at, and has since changed to using `allowedVersions`. It seems we're all slowly learning the lesson of how to correctly configure renovate. Ref https://github.com/cilium/hubble/pull/1239#issuecomment-1750266739

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -140,12 +140,7 @@
         "go",
         "golang"
       ],
-      "matchUpdateTypes": [
-        "patch",
-        "digest",
-        "pin",
-        "pinDigest"
-      ],
+      "allowedVersions": "<=1.20",
       "matchBaseBranches": [
         "v0.12"
       ],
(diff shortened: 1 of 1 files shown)


chancez:
@aanm mentioned we could also still use `matchUpdateTypes`, but only on major/minor and then set enabled: false, as it's done here: https://github.com/cilium/cilium/blob/main/.github/renovate.json5#L436-L447

The approach I've taken is a bit more explicit, and will require changing the Go/Alpine versions as we update the matchBaseBranches. The approach he's mentioned is less explicit, but basically would only require us to update the matchBaseBranches. Each has pros/cons. I'm open to suggestions on which way we would like to do it.

rolinh (approved):

lambdanis (approved):

PR #1263, opened 2023-10-19
Automerge Renovate minor/patch updates

chancez:
Renovate does a good job, if CI passes, let's auto merge the non-major updates.


glibsm (approved):
I think more might be required here, i'm not sure that renovate bot has access to merge onto main branch, especially without the required number of reviews. But lets see what happens, perhaps I'm wrong

chancez:
@glibsm yeah I agree, it's hard to tell until we try though.

Issue #952, opened 2023-03-21
Flows pers second implementation


glibsm:
This is more of a braindump rather than a concrete proposal.

## Context

Recent issue https://github.com/cilium/hubble/issues/911 got me thinking that perhaps sampling over the entire lifetime of the agent is not the way to go.

## Scenario

For example, lets assume the scenario where a cilium-agent has been running for 30days at roughly 100 flows/s.

If all of the sudden there is a 100x spike to 10k flows/s `hubble status` is going to report flows/s as 100, 101, 102... etc for a very long time until the 10k spike makes up for a sufficient percentage of the agent up-time.

## Problem

I would argue that a user checking `hubble status` should be able to see a more up-to-date picture of how much traffic has been flowing through cilium.

## Solution?

Perhaps on the server side we can keep a tally of how many flows were received over the last 1m or so and report that as flows/s to the clients in the API, rather then requiring them to divide the total number by uptime.

rolinh:
Isn't the use-case you report better served by metrics? At least, this is what I would monitor for an up-to-date picture of how much traffic has been flowing through Cilium.

PR #1272, opened 2023-10-23
Display server provided flows rate

glrf:
This change switches from reporting the calculate `Flows Per Second` field to the more up-to-date `flowsRate` field provided by the server.

If the server doesn't provide the flows rate, we fall back to calculating the flows per second using uptime and seen flows.

closes: #952


lambdanis (approved):
looks good

Issue #776, opened 2022-08-11
Support for filtering on HTTP headers


rolinh:
HTTP flows [contain headers](https://github.com/cilium/cilium/blob/862e191b3574e4a7c5d2046d28d980c163b50a6f/api/v1/flow/flow.proto#L456-L469) but Hubble doesn't support filtering flows based on HTTP headers. Using the CLI, we can already filter based on HTTP status code, methods and paths but filtering on headers is still missing.

gandro:
Are the HTTP headers even populated anymore? I'm unable to find the PR, but I think we disabled this in envoy for performance reasons.

dharmicksai:
Hello,
I would like to work on this issue, can I pick this up?

rolinh:
Hi @dharmicksai, certainly! I've assigned it to you, thanks!

rohithsai1904:
Hey @rolinh Can you assign it to me as well?

rolinh:
@dharmicksai Are you actively working on this or can @rohithsai1904 pick up the issue?

rolinh:
@rohithsai1904 Filtering is implemented in the backend, ie in Hubble server, not the CLI.
To implement support for filtering on HTTP headers, several changes need to happen:

1. Update the `FlowFilter` (see [here](https://github.com/cilium/cilium/blob/d4b41c1fd7bc2d9aef28857be3c89354e35b1a53/api/v1/flow/flow.proto#L373-L375)) so that a client can set HTTP header filters.
2. Implement the filter in Hubble (see [here](https://github.com/cilium/cilium/blob/master/pkg/hubble/filters/http.go)).
3. Update the Hubble CLI to add new flags to filter on HTTP headers.

rohithsai1904:
@rolinh Any idea how can i test the implementation i have written?

rolinh:
> @rolinh Any idea how can i test the implementation i have written?

Do you have a PR I can review? I'd start by implementing unit tests.

rohithsai1904:
I hv no PR as of now. Im working.

I updated flow filter in https://github.com/cilium/cilium/blob/d4b41c1fd7bc2d9aef28857be3c89354e35b1a53/api/v1/flow/flow.proto (IN MY LOCAL)

Do i have to update in https://github.com/cilium/cilium/blob/d4b41c1fd7bc2d9aef28857be3c89354e35b1a53/api/v1/flow/flow.pb.go as well?

rolinh:
[...]

PR #1277, opened 2023-10-28
[hubble observe]: Add http header filter

ChrsMark:
We are adding support for filtering on HTTP headers in https://github.com/cilium/cilium/pull/28851. This PR adds this feature in the CLI by adding a new flag `--http-header`

Fixes: https://github.com/cilium/hubble/issues/776

### How to test this manually

See https://github.com/cilium/cilium/pull/28851


rolinh:
@ChrsMark Do you need help with this work?

ChrsMark:
Hey @rolinh , I just wait for a pre-lease version of [cilium](https://pkg.go.dev/github.com/cilium/cilium?tab=versions) so as to update the [dependency](https://github.com/cilium/hubble/blob/2e199b4fef64d6169bad4c16086a81bf6b8b7bc8/go.mod#L6) and be able to use the types introduced with https://github.com/cilium/cilium/pull/28851.
If I miss sth here or there is another way to move this forward let me know.

rolinh:
Ah, I see. It's actually fine to upgrade the Cilium dep to whatever is `main` at the moment. Once Cilium v1.15 is branched (with `v1.15.0-rc.0`), we'll stick to vendoring the v1.15 branch until the next release of the CLI around Cilium v1.15.0.

cmd/observe/flows_filter.go
```diff
@@ -546,6 +547,13 @@ func (of *flowFilter) set(f *filterTracker, name, val string, track bool) error
 			f.HttpUrl = append(f.GetHttpUrl(), val)
 		})
 
+	case "http-header":
+		f.apply(func(f *flowpb.FlowFilter) {
+			h := strings.Split(val, ":")
```
lambdanis, line 552: The header value can contain a colon, so this should be `strings.SplitN(val, ":", 2)` I believe. Would be nice to have a test for such case too.

cmd/observe_help.txt
```diff
@@ -136,15 +137,15 @@ Server Flags:
       --request-timeout duration      Unary Request timeout. Only applies to non-streaming RPCs (ServerStatus, ListNodes, ListNamespaces). (default 12s)
       --server string                 Address of a Hubble server. Ignored when --input-file is provided. (default "localhost:4245")
       --timeout duration              Hubble server dialing timeout (default 5s)
-      --tls                           Specify that TLS must be used when establishing a connection to a Hubble server.
+      --tls                           Specify that TLS must be used when establishing a connection to a Hubble server.
```
lambdanis, line 140: :thinking: I fail to see any change, what happened here?

lambdanis (changes requested):
The CI failure looks legit - `TestTestHubbleObserve` is failing due to different ordering of CLI args.

cmd/observe_help.txt
ChrsMark: Should be some special chars (`^M`) most probably from my editor. I reverted those.

cmd/observe/flows_filter.go
ChrsMark: Nice catch! I think `SplitN` won't help here. I would just re-concatenate the result again except of the last item (that will be the value).

ChrsMark: I changed it like that and added a test-case header to cover it as well. Let me know what you think.

cmd/observe/flows_filter.go
```diff
@@ -546,6 +547,17 @@ func (of *flowFilter) set(f *filterTracker, name, val string, track bool) error
...
+			if len(h) > 0 {
+				hVal := h[len(h)-1]
+				key := strings.Join(h[:len(h)-1], ":")
```
tklauser, line 555: I think this should be the other way around. Everything before the first colon is the key and everything after is the value. I think `strings.SplitN` as mentioned by Anna would work as intended here. Another easier option might be [`strings.Cut`](https://pkg.go.dev/strings#Cut):

```go
key, hVal, _ := strings.Cut(h, ":")
```

tklauser (changes requested):

cmd/observe/flows_filter.go
```diff
@@ -546,6 +547,17 @@ func (of *flowFilter) set(f *filterTracker, name, val string, track bool) error
...
+				header := &flowpb.HTTPHeader{Key: key, Value: hVal}
```
tklauser, line 556: i.e.

```suggestion
			key, hVal, found := strings.Cut(val, ":")
			if found {
				header := &flowpb.HTTPHeader{Key: key, Value: hVal}
```

cmd/observe/flows_filter.go
ChrsMark: Got it! I doubled checked and only the values can contain `:`. Changed that.

cmd/observe/flows_filter.go
ChrsMark: Fixed properly after the correction at https://github.com/cilium/hubble/pull/1277#discussion_r1409417738

cmd/observe/flows_filter.go
```diff
@@ -546,6 +547,15 @@ func (of *flowFilter) set(f *filterTracker, name, val string, track bool) error
 			f.HttpUrl = append(f.GetHttpUrl(), val)
 		})
 
+	case "http-header":
+		f.apply(func(f *flowpb.FlowFilter) {
+			key, hVal, found := strings.Cut(val, ":")
+			if found {
+				header := &flowpb.HTTPHeader{Key: key, Value: hVal}
+				f.HttpHeader = append(f.GetHttpHeader(), header)
+			}
```
tklauser, line 556: Sorry, one more thing that just came to mind now. I think we should return an error in case no colon was found, i.e. an invalid HTTP header:

```suggestion
			if !found {
				return fmt.Errorf("invalid http-header value %q, expected name:value", val)
			}
			header := &flowpb.HTTPHeader{Key: key, Value: hVal}
			f.HttpHeader = append(f.GetHttpHeader(), header)

```

Would be nice extend `TestHTTPHeader` accordingly as well.

tklauser (changes requested):

cmd/observe/flows_filter.go
ChrsMark: Sure! I have added that.

lambdanis (approved):
Looks good to me, thanks for updates.

One detail: I would split the Cilium update into a separate commit.

tklauser:
> That vendor update related CI failure looks weird to me 🤔 : https://github.com/cilium/hubble/actions/runs/7049712881/job/19188814888?pr=1277#step:9:71
> 
> Is this sth that we should double check?

This is an issue that was recently fixed upstream (ref. https://github.com/cilium/cilium/pull/29481). Could you please update the vendored cilium version to latest `main` (`go get github.com/cilium/cilium@main && go mod tidy && go mod vendor`) in the commit bumping the dependency? Then this should be good to go!

ChrsMark:
Thank's @tklauser, it should be fine now :)

PR #1279, opened 2023-10-31
Make `--namespace` flag more intuitive and allow both `--namespace` and `--pod` flag

### LINKED ISSUE #359: --namespace is incompatible with --pod
Currently, it is not possible to give `--namespace` along with `--pod` to hubble.
```
% hubble observe --from-namespace cilium --from-pod hubble-relay
invalid argument "hubble-relay" for "--pod" flag: filters --from-namespace and --pod cannot be combined
```

An alternative way to express this is the following:
```
% hubble observe --from-label io.kubernetes.pod.namespace=cilium --from-pod hubble-relay
```

The issue with this is that `--from-pod hubble-relay` is implicitly referring to the `hubble-relay` pod _in the `default` namespace_. Here I would expect to see the the Flows from the `cilium/hubble-relay` pod.

cc @michi-covalent @gandro @nathanjsweet

glrf:
This PR essentially implements @kaworu's proposal in https://github.com/cilium/hubble/issues/359#issuecomment-702202613 .

This means instead of treating the `--namespace` flag as an alias for filtering for any pod in the namespace, we now treat the `--namespace` flag as a modifier for the `--pod` and `--service` flag. We also  try to catch namespace inconsistencies in flags and fail early.

There where a few edgecases that were not considered in the initial proposal:

* If we provide multiple namespace and/or pods/services we build the crossproduct of the provided values.
  
  For example `hubble observe -n foo -n bar --pod buzz` will return any flows from or to pods in namespace `foo` or `bar` named `buzz`

* If we don't provide any pod or service flags, i.e. `hubble observe -n foo`, we use the old behaviour of filtering for any pods in that namespace.

* We don't allow mixing `--namespace` and `--to-pod` or similar as it isn't clear what should happen in that case.

* We return an error if the namespace and the pod and/or service flags request flows from different namespace, as it is not clear what for example `hubble observe -n foo --pod bar/buzz --pod blub` should do.

This PR also doesn't implement the proposed `--all-namespaces` flag, as it requires server side changes. Will add this separately.

This technically introduces a breaking change. We no longer allow using both `--namespace` and `--from-service` or `--to-service`. 
Before this change `hubble observe --namespace foo --to-service buzz/bar` showed flows flows that either start at a pod in namespace `foo` and go to service `bar` in namespace `buzz`, or come from anywhere and go to service `buzz/bar` and a pod in namespace `foo` (which does not show any flows as long as `foo != buzz`)

If you want to replicate the old behavior you can use `hubble observe --pod foo/ --to-service buzz/bar`

closes #359

### DIFF
--- a/cmd/observe/flows_filter.go
+++ b/cmd/observe/flows_filter.go
@@ -27,6 +27,9 @@ type filterTracker struct {
 	// the resulting filter will be `left OR right`
 	left, right *flowpb.FlowFilter
 
+	// namespaces set through flags, will be applied to the pod and/or service
+	// filter when the flow filters are materialized.
+	ns, srcNs, dstNs namespaceModifier
 	// which values were touched by the user. This is important because all of
 	// the defaults need to be wiped the first time user touches a []string
 	// value.
@@ -74,21 +77,71 @@ func (f *filterTracker) applyRight(update func(*flowpb.FlowFilter)) {
 	update(f.right)
 }
 
-func (f *filterTracker) flowFilters() []*flowpb.FlowFilter {
-	if f.left == nil && f.right == nil {
-		return nil
-	} else if proto.Equal(f.left, f.right) {
-		return []*flowpb.FlowFilter{f.left}
+type namespaceModifier []string
+
+func (nm namespaceModifier) applySrc(ff *flowpb.FlowFilter) {
+	if len(ff.GetSourcePod()) == 0 && len(ff.GetSourceService()) == 0 {
+		ff.SourcePod = []string{""}
+	}
+	ff.SourcePod = addNamespacesToFilter(ff.GetSourcePod(), nm)
+	ff.SourceService = addNamespacesToFilter(ff.GetSourceService(), nm)
+}
+
+func (nm namespaceModifier) applyDest(ff *flowpb.FlowFilter) {
+	if len(ff.GetDestinationPod()) == 0 && len(ff.GetDestinationService()) == 0 {
+		ff.DestinationPod = []string{""}
+	}
+	ff.DestinationPod = addNamespacesToFilter(ff.GetDestinationPod(), nm)
+	ff.DestinationService = addNamespacesToFilter(ff.GetDestinationService(), nm)
+}
+
+func addNamespacesToFilter(filter []string, ns []string) []string {
+	if len(ns) == 0 || len(filter) == 0 {
+		return filter
 	}
 
-	filters := []*flowpb.FlowFilter{}
-	if f.left != nil {
-		filters = append(filters, f.left)
+	res := []string{}
+	for i := range filter {
+		if strings.Contains(filter[i], "/") {
+			res = append(res, filter[i])
+			continue
+		}
+		for j := range ns {
+			res = append(res, fmt.Sprintf("%s/%s", ns[j], filter[i]))
+		}
 	}
-	if f.right != nil {
-		filters = append(filters, f.right)
+	return res
+}
+
+func (f *filterTracker) flowFilters() []*flowpb.FlowFilter {
+	if f.left == nil && f.right == nil &&
+		len(f.ns) == 0 && len(f.dstNs) == 0 && len(f.srcNs) == 0 {
+		return nil
 	}
-	return filters
+	if f.left == nil {
+		f.left = &flowpb.FlowFilter{}
+	}
+	if f.right == nil {
+		f.right = &flowpb.FlowFilter{}
+	}
+
+	if len(f.dstNs) > 0 {
+		f.dstNs.applyDest(f.left)
+		f.dstNs.applyDest(f.right)
+	}
+	if len(f.srcNs) > 0 {
+		f.srcNs.applySrc(f.left)
+		f.srcNs.applySrc(f.right)
+	}
+	if len(f.ns) > 0 {
+		f.ns.applySrc(f.left)
+		f.ns.applyDest(f.right)
+	}
+
+	if proto.Equal(f.left, f.right) {
+		return []*flowpb.FlowFilter{f.left}
+	}
+	return []*flowpb.FlowFilter{f.left, f.right}
 }
 
 // Implements pflag.Value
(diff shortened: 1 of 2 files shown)


chancez (approved):
Overall LGTM. The tests seem thorough and make sense to me, which is the biggest part IMO.

cmd/observe/flows_filter.go
```diff
@@ -163,6 +241,60 @@ func (of *flowFilter) checkConflict(t *filterTracker) error {
...
+			t.srcNs.conflicts(t.right.SourceService),
+			t.dstNs.conflicts(t.right.DestinationPod),
+			t.dstNs.conflicts(t.right.DestinationService),
+			checkInconsistentNamespaces(t.right.SourcePod, t.right.SourceService),
+			checkInconsistentNamespaces(t.right.DestinationPod, t.right.DestinationService),
+		)
+	}
+
+	return errors.Join(errs...)
+
```
kaworu, line 295: nit
```suggestion
```

cmd/observe/flows_filter.go
```diff
@@ -163,6 +241,60 @@ func (of *flowFilter) checkConflict(t *filterTracker) error {
...
+	return nil
+}
+
+// checkNamespaceConflicts checks for conflicts in namespaces, pods and services
+func checkNamespaceConflicts(t *filterTracker) error {
+	var errs []error
+
+	if t.left != nil {
+		errs = append(errs,
+			t.ns.conflicts(t.left.SourcePod),
```
kaworu, line 267: trying to wrap my head around this check, specifically why we don't have `t.ns.conflicts(t.left.DestinationPod)` here.

Since `--{from,to}-pod` is incompatible with `--namespace` (i.e. the flag setting `t.ns`), then when `t.left.SourcePod` is non empty it was set through `--pod` and then `t.left.SourcePod == t.right.DestinationPod`, and both `t.left.DestinationPod` and `t.right.SourcePod` are empty right? If correct, I would be tempted to still check `t.ns.conflicts(t.left.DestinationPod)` (and respectively for svc and right) to make this codepath less reliant on understanding the different flag conflict / filter filling logic, since `conflicts()` will always return `nil` when given an empty slice, but I don't feel strongly about it.

cmd/observe/flows_filter.go
```diff
@@ -337,37 +469,38 @@ func (of *flowFilter) set(f *filterTracker, name, val string, track bool) error
 		})
 	// label filters
 	case "label":
+		if strings.HasPrefix(val, "k8s:io.kubernetes.pod.namespace=") {
```
kaworu, line 472: How about also handling the `io.kubernetes.pod.namespace=` prefix (i.e. without the `k8s:` part)? I guess most users don't provide the (cilium specific) [`k8s:` label source](https://github.com/cilium/cilium/blob/33573bd0e7e797e25d81ab183e742cb348729d95/pkg/labels/labels.go#L108-L109), and IMHO it's safe enough to assume that this label will and should only be defined by k8s. Thoughts?

kaworu (commented):
Awesome work @glrf :tada: Left a couple of comments but overall LGTM!

cmd/observe/flows_filter.go
glrf: Maybe I'm holding it wrong, but the `io.kubernetes.pod.namespace` label isn't set in my dev environment, all k8s labels have the `k8s:` prefix. So if I for example run `hubble observe --label io.kubernetes.pod.namespace=kube-system` I don't get any flows at all.

Should this actually work?

cmd/observe/flows_filter.go
glrf: Yes. My reasoning was that both `left.DestinationPod` and `right.SourcePod` should be empty. But you're right there is no good reason not to check it anyways.

cmd/observe/flows_filter.go
glrf: So looking into this I realized two things:

1. Yes, this should work but doesn't. There is a server side bug in the labels filter, where hubble doesn't properly handle labels with dots `.` and no prefix with a colon `:`. So that means `hubble observe --label k8s-app=hubble-relay`  works as expected but `hubble observe --label io.kubernetes.pod.namespace=kube-system` doesn't. The second one only matches a `io:kubernetes.pod.namespace` label. For now I'd assume that this works for this PR and will look into fixing this in `cilium/cilium`. 
2. My simplistic prefix label parsing is broken. If we try to use more complex label selector like `hubble observe --label k8s:io.kubernetes.pod.namespace=cilium,k8s:io.cilium in (is-awesome,rocks)`, we suddenly try to filter for pods in the namespace `cilium,k8s:io.cilium in (is-awesome,rocks)`. Doing this "properly" for more complex label selectors is actually not trivial

I'm starting to question if the whole special treatment for the namespace label actually makes sense. Doing this "properly" for any label selector IMO introduces a lot of complexity for not much gain and doing it for "simple" label selectors only (i.e. only if the selector is of the form `io.kubernetes.pod.namespace=<namespace>`) seems a bit magical and I think as a user it's not entirely clear for which selectors this will or won't work.

Lastly, who is actually going to write `hubble observe --label io.kubernetes.pod.namespace=kube-system get pod foo` instead of `hubble observe -n kube-system get pod foo` and is it worth it to special case this?

kaworu: @glrf all fair questions, let's drop the specific treatment of the namespace label!

kaworu (approved):

cmd/observe/flows_filter.go
```diff
@@ -108,8 +180,14 @@ type flowFilter struct {
...
-			{"from-fqdn", "from-ip", "from-namespace", "from-pod", "fqdn", "ip", "namespace", "pod"},
-			{"to-fqdn", "to-ip", "to-namespace", "to-pod", "fqdn", "ip", "namespace", "pod"},
+			{"from-fqdn", "from-ip", "ip", "fqdn", "from-namespace", "namespace"},
+			{"from-fqdn", "from-ip", "ip", "fqdn", "from-pod", "pod"},
+			{"to-fqdn", "to-ip", "ip", "fqdn", "to-namespace", "namespace"},
+			{"to-fqdn", "to-ip", "ip", "fqdn", "to-pod", "pod"},
+			{"to-pod", "namespace"},
+			{"from-pod", "namespace"},
+			{"to-service", "namespace"},
+			{"from-service", "namespace"},
```
kaworu, line 190: Seems to me that the patch doesn't introduce new incompatibilities, but lift of some (e.g. before the patch `--pod` and `--namespace` are incompatible, but ok after the patch). The [conflict test case coverage](https://github.com/cilium/hubble/blob/f65467be66e7db072e14ca202d2d7b1f4d56b5df/cmd/observe/flows_filter_test.go#L42-L56) is pretty low, so if correct it could be good to add a case for `namespace` and `pod`? If incorrect and new incompatibilities are added we should label the PR with `breaking-change`.

kaworu (changes requested):
Thanks @glrf patch LGTM but the commit message still reference the namespace label change that was dropped

> * We also treate `--label k8s:io.kubernetes.pod.namespace=foo` as a
  namespace flag

cmd/observe/flows_filter.go
glrf: There should not be any new incompatibilities, but I agree the the test coverage for this is very low and it's starting to get fairly complicated. I'll add a few test cases to be sure that this works as expected.

glrf: I added tests and realized we do add a new incompatibility. `hubble observe --namespace foo --to-service/--from-service bar` isn't valid anymore.

kaworu: Ok, marking the PR as a breaking change, please state this incompatibility in the PR's description so it will be easier to find back when making the releases notes :pray:

PR #1290, opened 2023-11-10
Add request timeout flag for unary RPCs

chancez:
Add the ability to configure the request timeout for unary RPCs (ServerStatus, ListNodes, ListNamespaces). Previously this was hardcoded at 12s and could not be overridden. I didn't implement this for streaming RPCs because timeouts for streams is significantly more nuanced.

Also updates a few cases where we're creating a cancellable context to cancel after the command func runs, but cobra provides a context for exactly that already, so use it.


rolinh (approved):
LGTM

PR #1309, opened 2023-12-04
Add --cluster flag for filtering by cluster

chancez:
This is just a short-cut to `--node-name <cluster-name>/` but should make it more easier for users to filter by cluster, since many users are not aware the cluster name is part of the node name field on a flow.

### DIFF
--- a/cmd/observe/flows.go
+++ b/cmd/observe/flows.go
@@ -348,6 +348,9 @@ func newFlowsCmdHelper(usage cmdUsage, vp *viper.Viper, ofilter *flowFilter) *co
 		"node-name", ofilter,
 		`Show all flows which match the given node names (e.g. "k8s*", "test-cluster/*.company.com")`))
 	filterFlags.Var(filterVar(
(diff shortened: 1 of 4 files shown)


kaworu (changes requested):
Patch LGTM, requesting change for a test `func` to be added for this new flag.

Also there is a potential incompatibility with the `node-name` flag, e.g.

```
hubble observe --cluster foo --node-name bar/cilium
```

Ideally we would check for this in the same fashion we do for e.g. the `--from-namespace` `--from-pod` since the "format" is the same (i.e. slash separated), but IMO non-blocking as `--node-name` is less useful than `--{from,to}-pod` and likewise for `--cluster` vs `--{from,to}-namespace`.

chancez:
Hmm, good point. I had considered making them conflict, but found they seemed fine when using both flags to filter by cluster, but I didn't think about filtering by cluster + node, so I can see the possible confusing behavior. I'll update the PR.

glrf (approved):

PR #1416, opened 2024-03-11
.github: Update renovate for v0.13 branches

chancez:
These were never updated to reflect the new base image versions in 0.13


rolinh (approved):
Thanks!

PR #1417, opened 2024-03-12
github: Run Docker build on all PRs

gandro:
This commit changes the "build-image-release.yaml" such that runs for every pull request and commit to the `main` branch. This ensures the build is tested before the actual release.

The credentials to push to quay.io are only available in the `release` deployment environment. This commit therefore changes the workflow to only use that environment if the workflow was triggered for a `v*` git tag. The `release` environment also needs approval from a Hubble maintainer, so the risk of accidentally pushing a release from a development branch is rather low.

### DIFF
--- a/.github/workflows/build-images-release.yaml
+++ b/.github/workflows/build-images-release.yaml
@@ -1,18 +1,43 @@
 name: Image Release Build
 
+# Any change in triggers needs to be reflected in the concurrency group.
 on:
   push:
+    branches:
+      - main
     tags:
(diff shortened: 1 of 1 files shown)


gandro:
I have manually tested the "proper" release workflow on my fork, which successfully pushed to quay: https://github.com/gandro/hubble/actions/runs/8250692155/job/22565930550

rolinh (approved):

viktor-kurchenko (approved):

PR #1465, opened 2024-04-22
Support --cel-expression filter in hubble observe

chancez:
This also updates cilium since the CEL filter expression is only available in main currently.


marseel (approved):
lgtm, but I would strongly encourage to try to get rid of `cilium/cilium` dependency until it's not too late :) 
Most of the imports seem to care only about `github.com/cilium/cilium/api` / `github.com/cilium/cilium/pkg/monitor/api` / `github.com/cilium/cilium/pkg/hubble/api/v1` and few other "random" imports.

glibsm:
> lgtm, but I would strongly encourage to try to get rid of `cilium/cilium` dependency until it's not too late :) Most of the imports seem to care only about `github.com/cilium/cilium/api` / `github.com/cilium/cilium/pkg/monitor/api` / `github.com/cilium/cilium/pkg/hubble/api/v1` and few other "random" imports.

We've been asking for a module in `api/` for years. There is no way to import api without importing cilium if they share the same module.

Also, the hubble codebase is about to merge into cilium/cilium anyway, so it's somewhat of a moot point now.

chancez:
Yeah, I'm going to create a PR to cilium/cilium now that https://github.com/cilium/cilium/pull/31893 is merged

PR #1490, opened 2024-05-13
Prepare for v0.13.4 release

gandro:


rolinh (changes requested):
I think we should bump Go to v1.21.10 before releasing, right? We're currently on v1.21.8.

rolinh:
Ah, I think #1398 broke renovate Go updates as it is now set to check go.mod but we never instruct renovate to update the Go version in the go.mod file. Not sure why the Dockerfile doesn't get updated however. :thinking: 
cc @chancez

PR #1522, opened 2024-06-05
Prepare for v1.16.0-pre.3 release

michi-covalent:


CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [v1.16.0-pre.3] - 2024-06-05
```
rolinh, line 8: I don't think we want to add pre-version changes to the changelog ahead of the final release.

michi-covalent: where should the go to? should we add `Pre-releases` section at the bottom of the file?

alternatively maybe we can just skip adding release notes for pre-releases altogether 🤔

rolinh: My thinking was:

- Update the changelog on patch/minor releases only (not when tagging a pre-version)
- Add the changes to every published release (that should be done automatically I believe, right?)

michi-covalent: > Update the changelog on patch/minor releases only (not when tagging a pre-version)

sounds good!

> Add the changes to every published release (that should be done automatically I believe, right?)

yeah we just need to specify the correct previous tag, which is mildly tricky when you are tagging `vX.Y.0` but we'll figure it out.

rolinh (approved):

PR #1548, opened 2024-07-11
build-images-release: Specify pattern when downloading image digests

learnitall:
This commit introduces a fix for the "Display Digests" job of the image build workflow. The job fails when trying to download image digest artifacts, which is caused by a recent change made to the docker/build-push-action that adjusted the way artifacts were stored. The fix is based off of
https://github.com/cilium/cilium/pull/33216, more details can be found there.


viktor-kurchenko (approved):

rolinh (approved):
Thanks!

PR #1600, opened 2024-10-06
Makefile cleanups / improvements

michi-covalent:
2 commits:

- drop building 32-bit release binaries
- use hubble target to build release binaries

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,8 @@ local-release: clean
 		for ARCH in $$ARCHS; do \
 			echo Building release binary for $$OS/$$ARCH...; \
 			test -d release/$$OS/$$ARCH|| mkdir -p release/$$OS/$$ARCH; \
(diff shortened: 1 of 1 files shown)


rolinh:
> Drop building 32-bit release binaries
> 
> Depending on the upstream Cilium packages you import, you may
> transitively pull in the agent, which does not build on 32-bit
> OSes due to the following error, which also breaks the Hubble build:
> 
> vendor/github.com/cilium/cilium/pkg/mountinfo/mountinfo_linux.go:56:27: invalid operation: fst.Type == mntType (mismatched types int32 and int64)

It looks like this error was fixed in https://github.com/cilium/cilium/commit/6f629af56b2a739b3a39d43d470822861100160e wasn't it? It's a bit unfortunate to drop support for 32-bit arm (I guess 386 is not so much used nowadays).

michi-covalent:
yeah the particular issue mentioned in the commit message has already been fixed i think. i can go either way regarding 32-bit binaries. fwiw we stopped releasing 32-bit binaries for cilium-cli a while back 💭 https://github.com/cilium/cilium-cli/pull/2136

chancez (approved):

PR #1631, opened 2024-12-10
Remove Dockerfile

michi-covalent:
Hubble CLI binary already gets shipped as a part of the Cilium Agent image, so let's not maintain yet another image.

- Remove Dockerfile and build-images-release.yaml.
- Put release.yml in the release GitHub environment.


Artyop (approved):

michi-covalent:
cc @cilium/hubble-maintainers what do you think about this change 👀

.github/workflows/release.yml
```diff
@@ -9,6 +9,7 @@ name: Create a release
 jobs:
   build:
     name: Create Release
+    environment: release
```
aanm, line 12: Was this intentional? :thinking:

aanm (changes requested):

glibsm (approved):

aanm (approved):

rolinh (approved):

PR #1778, opened 2026-04-21
.github: Add permissions for release workflow and support manually triggering it

chancez:
The latest release failed since we locked down the default permissions: https://github.com/cilium/hubble/actions/runs/24745109153/job/72394610612

Fix is to explicitly request permissions in the release workflow. Also make it possible to execute the release workflow against an existing tag so we can re-run it against the tag that failed.


joestringer (approved):
LGTM

PR #1788, opened 2026-06-01
renovate: add dependency cooldown

peoyekunle:
- adds a 5-day dependency cooldown to the Renovate config for stability and security to prevent supply chain vulnerabilities

- Vulnerability updates are exempt from this cooldown by default.


rolinh (approved):
Good idea, thanks!

